PR #7, opened 2020-03-10
l10n compatibility changes

shihaohong:
This updates `intl_en_US.arb` to `intl_en.arb` to use the American English translations as the fallback locale for the Gallery app. 

This fixes a regression caused by updates to the gen_l10n tool and will require a follow-up change to the Google translation script.


tool/l10n_cli/l10n_cli.dart
```diff
@@ -41,7 +41,7 @@ Future<void> englishArbsToXmls({bool isDryRun = false}) async {
   IOSink output =
       isDryRun ? stdout : File('$_l10nDir/intl_en_US.xml').openWrite();
```
guidezpl: Should this be changed to intl_en.xml too?

shihaohong: The reason this isn't changed to `intl_en.xml` is because the internal script expects `intl_en_US.xml`. It used to be `messages.xml` before it was `intl_en_US.xml`, but changing the name of the file confused the internal script and caused the internal system to no longer recognize the source file, resulting in a manual migration of existing translated messages. 

I'll add a new comment to the PR to advise against changing the name of the xml file for future developers.

shihaohong: Actually, the only place where the filename cannot be changed is internally, but keeping it as `intl_en_US.xml` on GitHub will keep the naming consistent.

PR #8, opened 2020-03-11
Use aspect ratio saved in Product to avoid layout shifting for Shrine

perclasson:
This fixes the issue of layout shifting when an item is added to the cart.

The root cause of this was that the image loading was not working the first time, thus the calculation was not the same as the second time. Now that we have the aspect ratio saved for every product, we can just use that in the calculation so it will work the same every time.

Closes https://github.com/material-components/material-components-flutter-gallery/issues/554

### DIFF
--- a/lib/studies/shrine/supplemental/balanced_layout.dart
+++ b/lib/studies/shrine/supplemental/balanced_layout.dart
@@ -226,7 +226,6 @@ List<List<Product>> balancedLayout({
   );
 
   // Check if this layout is cached.
-
   if (LayoutCache.of(context).containsKey(encodedParameters)) {
     return _generateLayout(
       products: products,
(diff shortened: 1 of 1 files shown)


lib/studies/shrine/supplemental/balanced_layout.dart
```diff
@@ -226,53 +207,17 @@ List<List<Product>> balancedLayout({
...
-
   final List<double> productHeights = [
-    for (final productSize in productSizes)
-      productSize.height /
-              productSize.width *
-              (largeImageWidth + smallImageWidth) /
-              2 +
+    for (final product in products)
+      ((1 / product.assetAspectRatio) *
+              ((largeImageWidth + smallImageWidth) / 2)) +
```
clocksmith: Some of these parens are not necessary:

1 / product.assetAspectRatio * productSize.width * (largeImageWidth + smallImageWidth) / 2

pennzht: @clocksmith 
I think it is `1 / product.assetAspectRatio * (largeImageWidth + smallImageWidth) / 2 + productCardAdditionalHeight`

PR #10, opened 2020-03-11
Dump translations

shihaohong:
First attempt at updating translations from google3


guidezpl:
Checked this out locally and looks like there's newlines in a couple ml strings which cause unterminated string errors when formatting. I've fixed it in the translation console and requested a new dump

PR #13, opened 2020-03-17
Add named routes for Gallery, to have URLs for each page in Flutter for Web

perclasson:
To make sure that we have URLs for each page in Flutter for Web, named routes were added for all of the pages. The most important changes can be seen inside of `lib/routes.dart`.

For each study we had a `NavigatorKey`, so that Focus support could be added to the back button. This was only used inside of Rally and the Starter app. After changes to the API in Flutter this no longer worked for Rally, and only inside the Starter app.

I decided to remove this as it did not work when you opened the same page multiple times (it caused an exception). We should implement Focus support for the back button when the framework *properly supports this* (see https://github.com/material-components/material-components-flutter-gallery/issues/491).

Closes https://github.com/material-components/material-components-flutter-gallery/issues/289


lib/studies/starter/home.dart
```diff
@@ -57,18 +48,16 @@ class HomePage extends StatelessWidget {
     if (isDesktop) {
       return Row(
         children: [
-          ListDrawer(
-            lastFocusNode: lastFocusNode,
-          ),
+          ListDrawer(),
```
guidezpl: nit: can be removed?

lib/routes.dart
```diff
@@ -0,0 +1,92 @@
...
+  /// [Navigator.pushNamed], the route name is matched with the [Path.pattern]
+  /// in the list below. As soon as there is a match, the associated builder
+  /// will be returned. This means that the paths higher up in the list will
+  /// take priority.
+  static List<Path> paths = [
+    Path(
+      r'^/demo/(?<slug>[\w-]+)$',
+      (context, matches) => DemoPage(slug: matches['slug']),
+    ),
+    Path(
```
guidezpl: If you're given a link to rally/login, does it still work?

lib/routes.dart
```diff
@@ -0,0 +1,92 @@
...
+      r'^/rally',
+      (context, matches) => StudyWrapper(study: RallyApp()),
+    ),
+    Path(
+      r'^/shrine',
```
guidezpl: nit: for cleanliness, I would create these strings using their definitions in the apps, e.g. RallyApp.defaultRoute

lib/pages/demo.dart
```diff
@@ -31,19 +31,31 @@ enum _DemoState {
...
-class DemoPage extends StatefulWidget {
-  const DemoPage({
+class DemoPage extends StatelessWidget {
+  const DemoPage({Key key, @required this.slug}) : super(key: key);
+
+  final String slug;
+
+  @override
+  Widget build(BuildContext context) {
+    return GalleryDemoPage(demo: slugToDemo(context)[slug]);
```
guidezpl: This might be slightly expensive to run on every build. Is there a way this can be refactored so this only happens once?

lib/routes.dart
perclasson: Yes, it works!

lib/routes.dart
perclasson: Sure. Due to to it being RegEx strings I cannot use string interpolation, so it'll be `r'/ + ShrineApp.defaultRoute`

lib/pages/demo.dart
perclasson: I agree, let me refactor this a bit.

PR #16, opened 2020-03-23
Run GitHub actions for pull requests

perclasson:
Run GitHub actions for pull requests


perclasson:
> LGTM but it doesn't look like `flutter format` is run in this.

Good call. I added it now.

PR #20, opened 2020-03-24
Add performance benchmarks (1/2)

guidezpl:
### Preparation for the perf transitions test
- Introduction of `GalleryDemoCategory` (study, material, cupertino, other)
   - Studies are now considered demos
  - Names, display text of demo and categories are centralised in `demos.dart`
- Demos are obtained by passing in a `GalleryLocalizations` object instead of a `BuildContext`
- Add `describe` getter to `GalleryDemo` to uniquely identify demos with the format 'demoName@demoCategory' (e.g. "Buttons@material")
- Add `ValueKey`s for integration tests to identify category headers, demo lists, and demos
- Create `test_driver/transitions_perf.dart`,  which launches the app in English and provides uniquely identifiable demo names to `transitions_perf_test.dart`

### Misc 
- Update flutter_localized_locales to 1.1.0 (which has a simpler loading model)
- Fix bug where demo categories on mobile were collapsing themselves when scrolled out of view
- Miscellaneous Podfile (iOS & macOS) updates
- Add Flutter Driver to  pubspec

## Future work
In a separate PR `test_driver/transitions_perf_test.dart `


lib/data/demos.dart
```diff
@@ -49,22 +50,56 @@ import 'package:gallery/themes/material_demo_theme_data.dart';
...
+
+  String displayTitle(GalleryLocalizations localizations) {
+    switch (this) {
+      case GalleryDemoCategory.material:
+        return 'MATERIAL';
+      case GalleryDemoCategory.cupertino:
+        return 'CUPERTINO';
+      case GalleryDemoCategory.other:
+        return localizations.homeCategoryReference;
+      case GalleryDemoCategory.study:
```
rami-a: Seems a bit odd for this case to do nothing, does Dart make you list it?

lib/data/demos.dart
```diff
@@ -49,22 +50,56 @@ import 'package:gallery/themes/material_demo_theme_data.dart';
...
   })  : assert(title != null),
-        assert(icon != null),
-        assert(configurations != null && configurations.isNotEmpty);
+        assert(category != null),
+        assert(subtitle != null),
+        assert(category == GalleryDemoCategory.study ||
+            (slug != null &&
+                icon != null &&
+                configurations != null &&
+                configurations.isNotEmpty));
```
rami-a: Formatting looks strange here but I guess we can't do much about this

lib/data/demos.dart
```diff
@@ -83,8 +118,43 @@ class GalleryDemoConfiguration {
...
-  final localizations = GalleryLocalizations.of(context);
+List<GalleryDemo> allGalleryDemos(GalleryLocalizations localizations) =>
+    studies(localizations).values.toList() +
+    materialDemos(localizations) +
+    cupertinoDemos(localizations) +
+    otherDemos(localizations);
+
+Map<Symbol, GalleryDemo> studies(GalleryLocalizations localizations) {
+  return <Symbol, GalleryDemo>{
+    #shrine: GalleryDemo(
```
rami-a: What is `Symbol`?

lib/data/demos.dart
clocksmith: Is this supposed to be empty?

lib/studies/fortnightly/app.dart
```diff
@@ -11,7 +11,7 @@ import 'package:gallery/layout/text_scale.dart';
 import 'package:gallery/l10n/gallery_localizations.dart';
 import 'package:gallery/studies/fortnightly/shared.dart';
 
-const fortnightlyTitle = 'Fortnightly';
+const _fortnightlyTitle = 'Fortnightly';
```
rami-a: Should this be removed and use the value from demos.dart instead?

lib/data/demos.dart
guidezpl: Yes, you have to list all enum members in switch clauses

guidezpl: Yes

lib/data/demos.dart
guidezpl: Yeah, I tried to get it to look better but didn't find something

guidezpl: Yeah, I tried to get it to look better but didn't find something

lib/data/demos.dart
guidezpl: I thought it was the Dart type for symbols but it's to be used for methods and members instead (https://api.flutter.dev/flutter/dart-core/Symbol/Symbol.html). I'll replace with strings

lib/studies/fortnightly/app.dart
guidezpl: It'd be a bit cumbersome to obtain so I'll leave it as is

lib/data/demos.dart
```diff
@@ -859,13 +960,15 @@ List<GalleryDemo> referenceDemos(BuildContext context) {
...
     ),
   ];
 }
 
 Map<String, GalleryDemo> slugToDemo(BuildContext context) {
+  final localizations = GalleryLocalizations.of(context);
   return LinkedHashMap<String, GalleryDemo>.fromIterable(
-    materialDemos(context) + cupertinoDemos(context) + referenceDemos(context),
+    allGalleryDemos(localizations),
     key: (dynamic demo) => demo.slug as String,
```
leafpetersen, line 972: @guidezpl I noticed in passing that this change looks like it adds a bunch of entries to the resulting map all with a `null` key.  That is, `allGalleryDemos` adds in the list of GalleryDemo entries produced from `studies`, each of which has `null` as the value of `demo.slug`.  The resulting map will have only a single entry for all of the `studies` entries, using `null` as the key.  Is this intentional?  Or have I somehow missed something here?

guidezpl: Studies were implemented first, and then demos added the concept of slugs, so I think that's why it's the way it is. AFAIK studies don't actually use `slugToDemo`

PR #22, opened 2020-03-25
Fix a method not found exception.

kangwang1988:
This issue is resulted from breaking changes in [flutter-localized-locales](https://github.com/guidezpl/flutter-localized-locales/commit/f7d1d69dd8a70fa3b91c3884a143f62342c58b74)
where allNativeNames is deleted.

Also ping @guidezpl , would it be better if flutter-localized-locales has tags so we can avoid issues like this.

### DIFF
--- a/lib/pages/settings.dart
+++ b/lib/pages/settings.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_localized_locales/flutter_localized_locales.dart';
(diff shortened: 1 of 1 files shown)


guidezpl:
Thanks for filing an issue. This shouldn't be a problem because dependencies are pinned https://github.com/flutter/gallery/blob/master/pubspec.lock#L133.

In any case, I have a PR (https://github.com/flutter/gallery/pull/20) fixing this so closing this issue.

kangwang1988:
@guidezpl When running the packages upgrade, the lock file would be useless, right?

PR #23, opened 2020-03-25
Use GoogleFonts.RobotoCondensed 

guidezpl:
Also fix weight (was too light), letter spacing for bodyText2 and padding for Rally headers to match spec

Now
Simulator Screen Shot - iPhone 11 Pro - 2020-03-25 at 16 37 22

Before
Screenshot 2020-03-25 at 16 52 07
Screenshot 2020-03-25 at 16 52 12


guidezpl:
Thanks!

PR #26, opened 2020-03-27
Disable animations on the web when navigating between pages

perclasson:
Disable animations on the web when navigating between pages

### DIFF
--- a/lib/routes.dart
+++ b/lib/routes.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:gallery/main.dart';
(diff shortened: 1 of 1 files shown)


lib/routes.dart
```diff
@@ -90,3 +97,26 @@ class RouteConfiguration {
     return null;
   }
 }
+
+class NoAnimationMaterialPageRoute<T> extends MaterialPageRoute<T> {
+  NoAnimationMaterialPageRoute({
+    @required WidgetBuilder builder,
+    RouteSettings settings,
+    bool maintainState = true,
```
clocksmith: Are these every changed? They can be passed to super without passing to this constructor.

perclasson: Done!

PR #27, opened 2020-03-29
Improve settings animation panel performance

clocksmith:
Improvements:
* Increases FPS by 7x for settings panel animation on web.
* Increases FPS by 2.5x for settings panel animation on mobile.

Changes:
* Remove unnecessary builds from backdrop and settings panel
* Move flare icon into its own AnimatedWidget
* rename frontPanel and backPanel to settings and home

before at full speed
before1

after at full speed
after1

before at timeDilation = 10
before10

after at timeDilation = 10
after10


lib/pages/backdrop.dart
```diff
@@ -208,44 +111,89 @@ class _BackdropState extends State<Backdrop>
...
+                  rect: _slideDownSettingsPageAnimation(constraints),
+                  child: settingsPage,
+                ),
+                // Slides the home page up and down below the bottom of the
+                // screen.
+                PositionedTransition(
+                  rect: _slideDownHomePageAnimation(constraints),
+                  child: homePage,
                 ),
               ],
```
perclasson: I tried it out on mobile and  the animation does not look exactly the same as before now, the staggering is changed.

Before:
animation_before

After:
animation_after

lib/pages/settings.dart
```diff
@@ -58,14 +52,14 @@ class _SettingsPageState extends State<SettingsPage> {
...
     super.initState();
 
-    // When closing settings, also shrink expanded setting.
-    widget.isSettingsOpenNotifier.addListener(() {
-      if (!widget.isSettingsOpenNotifier.value) {
-        setState(() {
-          expandedSettingId = null;
-        });
-      }
-    });
```
perclasson: If we now open the locale list for example in the settings, close the settings panel, and then reopen it the locale list will still be open. In the previous version, we closed the settings lists when the panel was closed.

Do we improve the performance by removing this?

lib/pages/backdrop.dart
```diff
@@ -166,35 +86,18 @@ class _BackdropState extends State<Backdrop>
...
-      widget.isSettingsOpenNotifier.value =
-          !widget.isSettingsOpenNotifier.value;
-    });
-    // Animate the settings icon.
+    _settingsPanelController.fling(
+        velocity: _isSettingsOpenNotifier.value ? -1 : 1);
+    _isSettingsOpenNotifier.value = !_isSettingsOpenNotifier.value;
     initAnimationLayer();
     isActive.value = true;
   }
```
perclasson: nit: Code like `initAnimationLayer`, `isActive.value = true`, `FlutterActorArtboard`, `FlareAnimationLayer`, `bool advance(FlutterActorArtboard artboard, double elapsed)` , `FlareController` are all related to the FlareIcon.

Not sure if its possible to move it down to the `SettingsIcon` or does that affect the performance?

lib/pages/backdrop.dart
```diff
@@ -362,17 +262,90 @@ class _BackdropState extends State<Backdrop>
...
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+const double _settingsButtonWidth = 64;
```
guidezpl: nit: would move these to the top

guidezpl:
`home_test.dart` needs to be updated

lib/pages/backdrop.dart
clocksmith: I tried that but did not have any luck, but I made a future issue for it: https://github.com/flutter/gallery/issues/76

lib/pages/settings.dart
clocksmith: Calling setState in the original way here was affecting performance (probably because setState is being called at the beginning of the position animation that is already going to call seState. I changed it to hide after the reverse animation completed instead, and this way the functionality is preserved while maintaining the performance when closing the panel.

PR #90, opened 2020-04-08
fix typography_demo body_text

mi6ock:
Fixed it because the description and style didn't match up.
In addition, the order was adjusted to the other.
Thank you!


mi6ock:
I'm so annoyed! Thank you so much!　@rami-a

Issue #73, opened 2020-03-30
Home screen on gallery needs improved a11y navigation


guidezpl:
## Acceptance Criteria
* Swipe traversal works in a sensible way
* The "crane" section is reachable via a11y navigation
* Expanding buttons/menus does not disrupt other lists/make announcements about being in a new list of more items than actually available.

## Reproduction steps

Run the gallery on a physical Android device with Talkback enabled.  Start swiping left/right to navigate around. Note:

- [x] Cards for demos don't scroll when activated
- [x] Cannot reach the Crane item at all
- [x] Expanding any of the buttons along the bottom announces in a list of 6 items
- [x] Swiping beyond those items leads to an invisible list (I think it's the menu list?)

/cc @willlarche @goderbauer @rami-a @darrenaustin @vick08 

Migrated from https://github.com/material-components/material-components-flutter-gallery/issues/134

guidezpl:
Next step is to integrate change from https://github.com/flutter/flutter/pull/45598

guidezpl:
Verified that the issues raised above are fixed. Found 2 remaining issues

- [ ] There are 2-3 unnecessary (and unannounced) items when opening the settings. Should go Close button  → Text scaling  → ...
- [ ] There is an unnecessary (and unannounced) item after the "Settings (button)" item

PR #95, opened 2020-04-10
Add implicit scrolling to carousel

guidezpl:
Related issue: #73


rami-a:
@clocksmith I believe this flag was added by @dnfield to enable better screenreader support for the carousel

guidezpl:
> @clocksmith I believe this flag was added by @dnfield to enable better screenreader support for the carousel

That's right

Issue #45, opened 2020-03-30
[Performance] Add flutter_gallery__transition_perf benchmarks of the new Flutter gallery app


guidezpl:
See https://github.com/flutter/samples/issues/309

Migrated from https://github.com/material-components/material-components-flutter-gallery/issues/559

liyuqian:
@guidezpl : do you have another issue that tracks monitoring new Flutter gallery app's performance in end-user devices (e.g., through https://firebase.google.com/docs/perf-mon), or is that effort also tracked under this issue?

rami-a:
I believe this is being tracked here:

https://github.com/flutter/gallery/issues/69

guidezpl:
https://github.com/flutter/flutter/pull/55181 closes this

PR #100, opened 2020-04-15
Add flutter_gallery__transition_perf benchmark

guidezpl:
This PR
- Adds a testMode flag to the app, which
    - expands demo categories on the home page mobile layout
    - disables feature highlight
- Adds a flutter_driver based test for transitions
    - Opens selected studies and demos, twice
    - Outputs `build/transitions.timeline_summary.json`
    - Supports both mobile and desktop layouts
    - iOS & Android fully supported (with a workaround)
        - `flutter drive  --profile --trace-startup -t test_driver/transitions_perf.dart -d ios`
    - macOS fully supported
        -  `flutter drive  --profile --trace-startup -t test_driver/transitions_perf.dart -d macos`

### Upcoming work
The benchmark will be partially runnable (I say 'will' because https://github.com/flutter/flutter/pull/54741 needs to land) on web  because the driver has trouble finding elements on the page.
- Unclear why this is happening, will follow up with the web team
- See https://flutter.dev/docs/cookbook/testing/integration/introduction#6b-web

### Related Issues
#45


lib/main.dart
```diff
@@ -17,10 +17,17 @@ import 'package:gallery/themes/gallery_theme_data.dart';
 
 void main() {
   GoogleFonts.config.allowHttp = false;
-  runApp(GalleryApp());
+  runApp(const GalleryApp(testMode: false));
```
clocksmith, line 20: The default value is false so it only needs to be set to true in the integration tests

test_driver/isolates_workaround.dart
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+    }
+    if (_streamSubscription != null) {
+      return;
+    }
+    _streamSubscription = _driver.serviceClient.onIsolateRunnable
+        .asBroadcastStream()
+        .listen((isolateRef) async {
+      final isolate = await isolateRef.load();
```
clocksmith, line 30: Extract these 7 lines into something like `_loadAndResumeIsolate()` since it's done here and above

test_driver/isolates_workaround.dart
```diff
@@ -0,0 +1,45 @@
...
+        if (log) {
+          print('Resuming isolate: ${isolate.numberAsString}:${isolate.name}');
+        }
+      }
+    });
+  }
+
+  Future<void> tearDown() async {
+    if (_streamSubscription != null) {
+      await _streamSubscription.cancel();
```
clocksmith, line 42: Should streamSubscription be set to null here too so tearDown doesn't try to cancel it more than once?

test_driver/transitions_perf.dart
```diff
@@ -5,21 +5,31 @@
...
-  return const JsonEncoder.withIndent('  ').convert(demoDescriptions);
+  switch (message) {
+    case 'demoDescriptions':
+      final demoDescriptions = allGalleryDemos(GalleryLocalizationsEn())
+          .map((demo) => demo.describe)
+          .toList();
+      return const JsonEncoder.withIndent('  ').convert(demoDescriptions);
+      break;
+    case 'isWeb':
+      return kIsWeb.toString();
```
clocksmith, line 25: How is handle messages used? Does it make sense that it can return 'True' or 'False'?

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,231 @@
...
+
+void main([List<String> args = const <String>[]]) {
+  group('flutter gallery transitions', () {
+    FlutterDriver driver;
+    IsolatesWorkaround workaround;
+
+    setUpAll(() async {
+      driver = await FlutterDriver.connect();
+
+      // See _handleMessages() in transitions_perf.dart.
```
clocksmith, line 182: I don't understand this comment. Where is _handleMessages used?

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,231 @@
...
+      _allDemos = List<String>.from(json.decode(
+        await driver.requestData('demoDescriptions'),
+      ) as List<dynamic>);
+      if (_allDemos.isEmpty) throw 'no demo names found';
+
+      // TODO: Remove workaround for https://github.com/flutter/flutter/issues/24703
+      final isWeb = await driver.requestData('isWeb') == 'true';
```
clocksmith, line 189: Ok I think this makes sense now. Maybe _handleMessasages could return an enum?

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,231 @@
...
+          await runDemos(_profiledDemos, driver);
+        },
+        streams: const <TimelineStream>[
+          TimelineStream.dart,
+          TimelineStream.embedder,
+        ],
+      );
+
+      final summary = TimelineSummary.summarize(timeline);
+      await summary.writeSummaryToFile('transitions', pretty: true);
```
clocksmith, line 223: How are we going to consume these summaries after the tests complete?

test_driver/transitions_perf.dart
guidezpl: It serves as the channel to pass messages from the app with the driver. It can only pass strings, though. I had to do it this way because only the app has access to things like kIsWeb

test_driver/transitions_perf_test.dart
guidezpl: Unfortunately, can only use String https://api.flutter.dev/flutter/flutter_driver_extension/DataHandler.html

test_driver/transitions_perf_test.dart
guidezpl: That is yet to be determined

test_driver/isolates_workaround.dart
guidezpl: I don't think so because it seems to only be called once through flutter_driver's `tearDownAll`

test_driver/transitions_perf_test.dart
clocksmith: Ah, this is good then

lib/data/gallery_options.dart
```diff
@@ -52,6 +53,7 @@ class GalleryOptions {
   final Locale _locale;
   final double timeDilation;
   final TargetPlatform platform;
+  final bool testMode; // True for integration tests.
```
rami-a, line 56: I think the name of this could be something like `isTestMode`, just `testMode` makes it sound like an enum.

test_driver/transitions_perf.dart
```diff
@@ -5,21 +5,31 @@
 import 'dart:async';
 import 'dart:convert' show JsonEncoder;
 
+import 'package:flutter/foundation.dart' show kIsWeb;
+import 'package:flutter/material.dart';
 import 'package:flutter_driver/driver_extension.dart';
 import 'package:gallery/data/demos.dart';
+import 'package:gallery/main.dart' show GalleryApp;
+
 import 'package:gallery/l10n/gallery_localizations_en.dart';
```
rami-a, line 14: This should be back above the main import to remain alphabetical

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,231 @@
...
+// These names must match the output of GalleryDemo.describe in
+// lib/data/demos.dart.
+const List<String> _profiledDemos = <String>[
+  'Shrine@study',
+  'Rally@study',
+  'Crane@study',
+  'Fortnightly@study',
+  'Bottom navigation@material',
+  'Buttons@material',
+  'Cards@material',
```
rami-a, line 31: Should we also profile one of the cupertino demos?

guidezpl: I agree, I'll add one

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,232 @@
...
+    await driver.scrollUntilVisible(
+      demoList,
+      demoItem,
+      dxScroll: -500,
+      dyScroll: -50,
+      alignment: 0.5,
+      timeout: const Duration(seconds: 10),
+    );
+
+    for (var i = 0; i < 2; i += 1) {
```
clocksmith, line 155: Any reason that this is done exactly 2 times? Maybe add a comment explaining why

test_driver/transitions_perf_test.dart
```diff
@@ -0,0 +1,232 @@
...
+
+/// Scrolls each each demo into view, launches it, then returns to the
+/// home screen, twice.
+Future<void> runDemos(List<String> demos, FlutterDriver driver) async {
+  String currentDemoCategory;
+  SerializableFinder demoList;
+  SerializableFinder demoItem;
+
+  for (final demo in demos) {
+    if (_skippedDemos.contains(demo)) continue;
```
clocksmith, line 116: Could instead iterate over demos.where((demo) => !skippedDemos.contains(demo))

PR #105, opened 2020-04-16
Add integration test for compiled js bundle size

perclasson:
Closes https://github.com/flutter/gallery/issues/101

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
       if: runner.os == 'macOS'
 
   integration-test:
-    name: Integration test
+    name: Integration tests
     runs-on: ubuntu-latest
     steps:
     # Set up Flutter and add it to the path.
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -62,3 +62,35 @@ jobs:
...
+    # Clone the Gallery repository under `code`, to avoid conflicts with `flutter`.
+    - name: Clone the Gallery repository.
+      uses: actions/checkout@v2
+      with:
+        path: 'code'
+
+    - name: Get packages for the Flutter project.
+      run: flutter pub get
+      working-directory: code
+    - name: Run Flutter integration tests.
```
guidezpl, line 94: Could you rename this to 'Run gallery_compile_test' with a command for that test specifically? That way when we add more tests it'll be easier to view their outputs separately in the Github run result.

.github/workflows/main.yml
```diff
@@ -62,3 +62,35 @@ jobs:
       # TODO: Enable verify-code-segments for Unix and Windows when it behaves the same as on macOS:
       # https://github.com/material-components/material-components-flutter-gallery/issues/565
       if: runner.os == 'macOS'
+
+  integration-test:
+    name: Integration test
```
guidezpl, line 67: nit: Integration tests

.github/workflows/main.yml
perclasson: That makes sense, will change it to that

.github/workflows/main.yml
perclasson: Done!

Issue #99, opened 2020-04-14
[Performance] Crane tab jank


willlarche:
**Describe the bug**
Going between tabs in Crane shows jank on the photos that are sliding.

guidezpl:
@willlarche I'm unable to get below 20fps with the latest master, can you repro?

willlarche:
I don't know if it's below 20 but it's still way below 60. Is 20 the goal?

guidezpl:
60 is the goal, but for something like this, I think it's acceptable so long as it isn't noticeable. Do you see a difference between when you reported the bug and now?

willlarche:
I can’t recall. But I do think it’s noticeable enough to deserve a look.

guidezpl:
Fixing https://github.com/flutter/flutter/issues/21389 would help

guidezpl:
https://github.com/flutter/gallery/issues/354 is now believed to be the primary blocker

PR #114, opened 2020-04-21
Only render Crane visible items

guidezpl:
- Move state (destination list) for a given tab to _FrontLayer
- Use StaggeredGridView from flutter_staggered_grid_view to create the 4, 2 or 1 column layout
    - Added support for 2 column layout
- Only render Destination cards visible above the fold
- Fix padding issue for form headers
- Fix image cropping issue on mobile 

Screenshot 2020-04-21 at 19 27 24

Related #99


lib/studies/crane/backdrop.dart
```diff
@@ -54,18 +93,27 @@ class _FrontLayer extends StatelessWidget {
...
+              if (index == 0) {
+                return _header();
+              } else {
+                return DestinationCard(destination: destinations[index]);
+              }
+            },
+            staggeredTileBuilder: (index) => index == 0
+                ? StaggeredTile.fit(crossAxisCount)
+                : const StaggeredTile.fit(1),
+            itemCount: destinations.length,
```
rami-a, line 116: Shouldn't this be +1 for the header?

lib/studies/crane/header_form.dart
```diff
@@ -11,11 +11,13 @@ const appPaddingLarge = 120.0;
...
 
 class HeaderFormField {
+  final int index;
   final IconData iconData;
   final String title;
   final TextEditingController textController;
 
-  const HeaderFormField({this.iconData, this.title, this.textController});
+  const HeaderFormField(
+      {this.index, this.iconData, this.title, this.textController});
```
rami-a, line 20: Nit: trailing comma

lib/studies/crane/item_cards.dart
```diff
@@ -8,114 +8,99 @@ import 'package:flutter/rendering.dart';
...
+      highlightColor: Colors.red.withOpacity(0.5),
+      onPressed: () {},
+      child: card,
+    );
   }
 }
 
-class _DestinationCard extends StatelessWidget {
-  _DestinationCard({@required this.destination}) : assert(destination != null);
+class DestinationImage extends StatelessWidget {
```
rami-a, line 88: Make private I think

lib/studies/crane/backdrop.dart
guidezpl: Good find!

guidezpl: It didn't work well, with it, not sure why, but keeping as is

guidezpl:
> LGTM, does this improve framerate when switching tabs?

It's a slight improvement as the very least, moderate if the screen is smaller and fewer destination cards have to load.

PR #121, opened 2020-04-24
Slider flag not needed

JoseAlba:
Slider Feature Flag defaults true. So this flag is not necessary anymore.


JoseAlba:
Waiting for https://github.com/flutter/flutter/pull/55599 to land

Issue #87, opened 2020-04-07
Reduce repo size


guidezpl:
<img width="500" alt="Screenshot 2020-04-24 at 12 48 51" src="https://user-images.githubusercontent.com/6655696/80206017-eca25a00-862b-11ea-997a-eb41e3c3cce9.png">
The repo has became unreasonably large (352mb), mostly due to a repo history (285mb) containing now deleted media files.

## Success criteria

- [x]  Move crane and splash_effects assets to their own package (and maybe fortnightly for consistency).
- [ ] Rewrite the history with a tool like BFS (see https://blog.jessitron.com/2013/08/18/finding-and-removing-large-files-in-git-history/) to remove large files
- [x] Delete `gallery_localizations.dart`, which is no longer needed as the translations are in individual `.dart` files now

guidezpl:
BFG history rewrite complete. Starting fresh with a
 ```git clone git@github.com:flutter/gallery.git``` (SSH)
```git clone https://github.com/flutter/gallery.git``` (HTTPS)
is strongly recommended.

## Results
### As a bare repository (252mb  → 47mb)

Before | After
------------ | -------------
<img width="991" alt="Screenshot 2020-06-27 at 23 22 33" src="https://user-images.githubusercontent.com/6655696/85932920-4431a000-b8d1-11ea-86e4-90a4288580db.png"> |  <img width="991" alt="Screenshot 2020-06-27 at 23 43 42" src="https://user-images.githubusercontent.com/6655696/85932927-4f84cb80-b8d1-11ea-8799-361a3efe1f5f.png">

### Replacing an existing repository (477mb  → 67mb)

Before | After
------------ | -------------
<img width="1488" alt="Screenshot 2020-06-27 at 23 21 45" src="https://user-images.githubusercontent.com/6655696/85932960-b609e980-b8d1-11ea-98db-1a0e5f8d0fd5.png"> | <img width="1488" alt="Screenshot 2020-06-27 at 23 49 58" src="https://user-images.githubusercontent.com/6655696/85932961-ba360700-b8d1-11ea-8b6a-6d31ab3b3a9f.png">

### Zipped
Now 9.7mb

PR #125, opened 2020-04-27
Move study and splash effects assets to package

guidezpl:
Part of work for #87 

Updated https://pub.dev/packages/flutter_gallery_assets/versions/0.2.0


pubspec.yaml
```diff
@@ -43,18 +43,59 @@ flutter:
...
     - assets/icons/settings/
     - assets/logo/
-    - assets/splash_effects/
-    - packages/rally_assets/logo.png
-    - packages/rally_assets/thumb.png
-    - assets/crane/logo/
-    - assets/crane/destinations/
-    - assets/fortnightly/
+    - assets/studies/
+    - packages/flutter_gallery_assets/crane/destinations/eat_1.jpg
```
rami-a, line 52: It's annoying that when you use a package, you have to specify each asset individually. But oh well

pubspec.yaml
```diff
@@ -24,7 +24,7 @@ dependencies:
   url_launcher: ^5.1.6
   shared_preferences: ^0.5.4
   collection: ^1.14.0
-  flutter_gallery_assets: ^0.1.9
+  flutter_gallery_assets: ^0.2.0
```
rami-a, line 27: Wouldn't the old assets still exist in the git history of the repo, so would the repo size actually be reduced?

pubspec.yaml
guidezpl: I'm working on a change to allow the same directory specifying as in local assets, just need more time to get it done :)

pubspec.yaml
guidezpl: My next step as part of #87 is to rewrite the history to scrap all deleted files

Issue #39, opened 2020-03-30
Add Screenshot tests


perclasson:
## Acceptance Criteria
* Golden tests for the demos
* ~~Add the tests to https://github.com/flutter/tests/blob/master/registry/flutter_gallery.test~~ This is not possible due to differences in how the tests will be run on different machines.

Migrated from https://github.com/material-components/material-components-flutter-gallery/issues/547

PR #127, opened 2020-04-28
Add golden tests for demo page

perclasson:
Closes https://github.com/flutter/gallery/issues/39


golden_test/testing/font_loader.dart
```diff
@@ -35,10 +36,29 @@ Future<Map<String, List<Future<ByteData>>>> loadFontsFromManifest() async {
   return fontFamilyToData;
 }
 
+Map<String, List<Future<ByteData>>> loadFontsFromTestingDir() {
```
rami-a, line 39: Why do all the roboto font files have to be included now?

golden_test/demo_test.dart
```diff
@@ -0,0 +1,85 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+import 'package:gallery/main.dart';
+
+import 'testing/precache_images.dart';
+import 'testing/util.dart';
+
+const demoBannerRoute = '/demo/banner';
+
+void main() {
```
rami-a, line 15: As discussed offline, rename all these tests to not specify banner but instead be more generic for demo page

golden_test/testing/font_loader.dart
perclasson: We were not using the Roboto font on the home page, but it is on the demo page, so I never noticed that we needed to load it as well.

In the previous code we were only loading the google fonts that were added and the icons from the `pubspec.yaml`. As the Roboto font is added by default in Flutter and not added in the `pubspec.yaml` or in the `google_fonts` folder, we need to explicitly load it here in the tests.

golden_test/demo_test.dart
perclasson: Done!

PR #129, opened 2020-04-28
Add new files from `flutter create .`

clocksmith:
This includes manifest.json which helps with PWA compatibility.


web/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "gallery",
+    "short_name": "gallery",
+    "start_url": ".",
+    "display": "minimal-ui",
+    "background_color": "#0175C2",
+    "theme_color": "#0175C2",
+    "description": "A new Flutter project.",
```
rami-a, line 8: should we update some of these values?

web/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+    "orientation": "portrait-primary",
+    "prefer_related_applications": false,
+    "icons": [
+        {
+            "src": "icons/Icon-192.png",
```
rami-a, line 13: Not necessary for this PR, but we should update all of these new assets to be the proper ones

clocksmith: OK, what are the proper ones?

web/manifest.json
clocksmith: Done

web/manifest.json
guidezpl: Let me get you some images

web/manifest.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "gallery",
+    "short_name": "gallery",
```
guidezpl, line 3: nit: Flutter Gallery

web/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+    "start_url": ".",
+    "display": "minimal-ui",
+    "background_color": "#0175C2",
+    "theme_color": "#0175C2",
```
guidezpl, line 7: 451B6F

clocksmith: Done

web/manifest.json
clocksmith: Done

PR #130, opened 2020-04-28
Update index.html with a new `flutter create` that links manifest.json and loads flutter_service_worker.js

clocksmith:


web/index.html
```diff
@@ -2,14 +2,37 @@
 <html>
 <head>
   <meta charset="UTF-8">
+  <meta content="IE=Edge" http-equiv="X-UA-Compatible">
+  <meta name="description" content="A new Flutter project.">
```
yjbanov, line 6: This should probably have a better description.

clocksmith: I have it fixed in next PR which also fixes favicons: https://github.com/flutter/gallery/pull/131

PR #137, opened 2020-04-29
Add Google Analytics

kevmoo:
Closes https://github.com/flutter/gallery/issues/135

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -3,29 +3,25 @@
 <head>
   <script async src="https://www.googletagmanager.com/gtag/js?id=UA-67589403-10"></script>
   <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-      gtag('config', 'UA-67589403-10');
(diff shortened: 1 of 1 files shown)


web/index.html
```diff
@@ -1,6 +1,32 @@
 <!DOCTYPE html>
 <html>
 <head>
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-67589403-10"></script>
+  <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-67589403-10');
+      window.onerror = function (msg, url, lineNo, columnNo, error) {
```
ditman, line 10: Maybe use [`window.addEventListener('error', function)`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#window.addEventListenererror) instead of overwriting `window.onerror`?

The only difference with your code is that instead of receiving separate params with the error information, you get an ErrorEvent [object](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent).

kevmoo: ack!

Issue #119, opened 2020-04-24
Home page should not build non-visible items


guidezpl:
- [x] Use a ListView.builder instead of a ListView for Material, Cupertino, and Other demos

PR #146, opened 2020-05-06
Use ListView.builder where possible

guidezpl:
Closes https://github.com/flutter/gallery/issues/119

Removes padding on top of the desktop demo category lists


lib/pages/home.dart
```diff
@@ -421,22 +421,12 @@ class _DesktopCategoryItem extends StatelessWidget {
...
               Flexible(
-                // Remove ListView top padding as it is already accounted for.
-                child: MediaQuery.removePadding(
-                  removeTop: true,
-                  context: context,
-                  child: ListView(
-                    // Makes integration tests possible.
-                    key: ValueKey('${category.name}DemoList'),
-                    children: [
-                      const SizedBox(height: 12),
```
perclasson, line 432: I wonder why we had this to begin with. It doesn't look that great so makes sense to remove it.

rami-a: It was in the designs, but I agree that it looks better without it

PR #151, opened 2020-05-07
Add performance testing to Gallery on Web

### LINKED ISSUE #103: [Performance] Add performance testing to Gallery on Web
Add tests to collect frame-related metrics (frame build time, frame rasterization time) for Gallery on Web.

See also: https://github.com/flutter/gallery/issues/192

pennzht:
## Overview
This PR adds performance testing to Gallery on Web.

The new tests will:
* Run Gallery in a Chrome browser;
* Automatically tap into and out of demos (studies, widget demos);
* Measures frame-related metrics during this period.

See `devicelab/README.md` on how to run.

Because studies are more complicated than widget demos, and some widget demos contain animations of their own, demos are split into three categories, and their performance are measured separately.

## Related issues
#103


yjbanov:
Why not simply define a new task in the flutter/flutter's version of the devicelab? This copy won't automatically run in the devicelab and it duplicates a lot of code.

pennzht:
@yjbanov 

> Why not simply define a new task in the flutter/flutter's version of the devicelab? This copy won't automatically run in the devicelab and it duplicates a lot of code.

I will add a task in flutter/flutter too, which will clone this repo and run the task automatically — similar to [new_gallery__transition_perf.dart](https://github.com/flutter/flutter/blob/master/dev/devicelab/bin/tasks/new_gallery__transition_perf.dart)
I wrote this test in flutter/gallery because it imports many files from the gallery.

yjbanov:
> @yjbanov
> 
> > Why not simply define a new task in the flutter/flutter's version of the devicelab? This copy won't automatically run in the devicelab and it duplicates a lot of code.
> 
> I will add a task in flutter/flutter too, which will clone this repo and run the task automatically — similar to [new_gallery__transition_perf.dart](https://github.com/flutter/flutter/blob/master/dev/devicelab/bin/tasks/new_gallery__transition_perf.dart)

The devicelab expects to execute its own code to run benchmarks, not a copy of itself. So I expect this will lead to difficulties.

> I wrote this test in flutter/gallery because it imports many files from the gallery.

It's totally OK to keep the test in the gallery. I'm mostly referring to code under `devicelab/`. That code does not import `package:gallery` AFAICT. I think it can be removed from here. All you need is the simplest way to run the test as a presubmit step, but I expect that should be no more than 100 lines of code to implement. Copying thousands of lines of devicelab code seems like an overkill.

I also think it's OK to import the gallery code from flutter/flutter. We could have a [git dependency](https://dart.dev/tools/pub/dependencies#git-packages) that we'd bump once in a while (or have a roller process that does it automatically). I'm not saying that it's necessary for this use-case, just that it's not wrong to import from gallery, if need be.

Anyway, I'm leaving the final decision whether you need a copy of devicelab to your team (/cc @clocksmith @guidezpl @perclasson). I will review other parts of the PR though.

lib/benchmarks/scroll.dart
```diff
@@ -0,0 +1,96 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
yjbanov, line 1: Have you considered using https://api.flutter.dev/flutter/flutter_test/WidgetController-class.html to manipulate widgets programmatically (e.g. you can use `drag` to scroll).

lib/benchmarks/benchmark_controller.dart
```diff
@@ -0,0 +1,411 @@
...
+import 'dart:html' as html;
+import 'dart:math' as math;
+
+import 'gallery_recorder.dart';
+import 'recorder.dart';
+
+typedef RecorderFactory = Recorder Function();
+
+const bool isCanvasKit =
+    bool.fromEnvironment('FLUTTER_WEB_USE_SKIA', defaultValue: false);
```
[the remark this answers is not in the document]
pennzht: Removed.
Thank you!

lib/benchmarks/scroll.dart
pennzht: `scroll.dart` contains a few helper functions for scrolling, such as `scrollUntilVisible`, which controls whether the scroll is animated, and how the target widget is aligned with the viewport. It's not very complicated, so I think we can keep it here.

PR #153, opened 2020-05-08
Remove version requirement 

perclasson:
Remove version requirement as its higher than v1.16 anyways and bound to master.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,7 +5,6 @@ version: 2.3.0+020300
 
 environment:
   sdk: ">=2.6.0 <3.0.0"
-  flutter: ^1.16.0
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -5,7 +5,6 @@ version: 2.3.0+020300
 
 environment:
   sdk: ">=2.6.0 <3.0.0"
-  flutter: ^1.16.0
```
guidezpl, line 8: I think this should still be specified

perclasson: I think it does not make sense as we're bound to master. So it's not actually accurate. If we could set this to be master then it would make sense.

PR #160, opened 2020-05-14
Automatic iOS signing instead of the manual one

liyuqian:
This would allow running the app on an iPhone without extra Xcode
setup, and therefore allows the device lab to automatically benchmark
the app on iOS.


liyuqian:
@gaaclarke @xster : I tested this change locally. Without it, `flutter drive` would complain about signing. I'm not very familiar with the `.pbxproj` file so I just copied the setup from the old gallery's `.pbxproj` file: https://github.com/flutter/flutter/blob/master/dev/integration_tests/flutter_gallery/ios/Runner.xcodeproj/project.pbxproj. Please let me know if this is the right thing to do.

PR #163, opened 2020-05-19
Add Web Benchmarks for Flutter Gallery (Gallery Side)

pennzht:
## Description
This PR, together with the PR on flutter/flutter, measures frame-related metrics for the Flutter Gallery on Chrome.

This PR also:
* Uses a package for all assets under `assets/` and `fonts/`, so that the gallery can be imported into `macrobenchmarks` in `flutter/flutter`. See also: https://github.com/flutter/flutter_gallery_assets/pull/39
* Updates demo IDs to use unique strings that do not depend on locales.

## Related Issues
Together with the [PR on Flutter](https://github.com/flutter/flutter/pull/57576), will close the issue https://github.com/flutter/gallery/issues/103


lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  /// Opens and quits demos that are specified by [runCriterion], twice.
+  Future<void> automateDemoGestures() async {
+    await animationStops();
+
+    controller = LiveWidgetController(WidgetsBinding.instance);
+
+    reporter('==== List of demos to be run ====');
+    for (final demo in demoNames) {
+      if (!runCriterion(demo)) continue;
```
JoseAlba, line 87: I'm under the impression that using the continue method is bad practice but not too sure

lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,199 @@
...
+      await scrollUntilVisible(
+        element: demoButton,
+        animated: false,
+        strict: true,
+      );
+
+      // Skip demo if it does not pass `runCriterion`.
+      // This continue statement is placed here because we need to scroll
+      // through each `Scrollable` to find the `demoButton`.
+      if (!runCriterion(demo)) continue;
```
JoseAlba, line 117: Similar to note above

lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,199 @@
...
+
+      reporter('Running demo "$demo"');
+
+      for (var i = 0; i < 2; ++i) {
```
JoseAlba, line 121: Where does two come from and if this is always going to be doesn't this only run one time and doesn't necessarily have to be used

lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,199 @@
...
+      final atSymbolIndex = demo.lastIndexOf('@');
+      if (atSymbolIndex < 0) {
+        return '';
+      } else {
+        return demo.substring(atSymbolIndex + 1);
+      }
+    }
+
+    // Select a demo from each category, so that we can test scrolling each one.
+    final coveredCategories = <String>{};
```
JoseAlba, line 156: final => final String

lib/benchmarks/gallery_automator.dart
pennzht: We open and close every demo twice, in order to measure initial and subsequent builds.
Similar to https://github.com/flutter/gallery/blob/master/test_driver/transitions_perf_test.dart#L157

lib/benchmarks/gallery_automator.dart
pennzht: The type is `Set<String>`.
The Gallery repo also requires omitting local variable types, as shown in this `flutter analyze` report.

```
   info • Omit type annotations for local variables • lib/benchmarks/gallery_automator.dart:156:5 •
          omit_local_variable_types
```

lib/benchmarks/README.md
```diff
@@ -0,0 +1,10 @@
+# Devicelab Tests
+
+This directory, `lib/benchmarks/`, is used for performance testing for the Gallery on Web.
```
guidezpl, line 3: I think the benchmarks directory would be better placed at the top level, keeping lib for the app source.

pennzht: I think I can only put it inside `lib`.

The benchmark is not run from inside this folder, but from `macrobenchmarks` inside `flutter/flutter`; please see https://github.com/flutter/flutter/pull/57576

`macrobenchmarks` needs to import the gallery and use the `GalleryAutomator` class, so this file needs to stay inside `lib`.

lib/benchmarks/gallery_automator.dart
pennzht: Changed to block statement, which is easier to read.

lib/benchmarks/gallery_automator.dart
pennzht: Changed to block statement, which is easier to read.

lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,201 @@
...
+    this.testScrollsOnly = false,
+  });
+
+  final String benchmarkName;
+  final void Function(String) reporter;
+  final Future<void> Function(dynamic error, StackTrace stackTrace) reportError;
+  final bool Function(String) runCriterion;
+  final bool testScrollsOnly;
+
+  bool finished = false;
```
clocksmith, line 58: Where is this consumed?

lib/benchmarks/scroll.dart
```diff
@@ -0,0 +1,125 @@
...
+    (renderObject as RenderBox).localToGlobal(Offset.zero);
+
+Rect _absoluteRect(RenderObject renderObject) =>
+    _absoluteTopLeft(renderObject) & renderObject.paintBounds.size;
+
+Size _windowSize(BuildContext context) => MediaQuery.of(context).size;
+
+Rect _windowRect(BuildContext context) => Offset.zero & _windowSize(context);
+
+bool _isSuperset({@required Rect large, @required Rect small}) =>
```
clocksmith, line 26: This could also be `large.intersect(small) == small`, but maybe what you have here is more clear

pennzht: I agree that `large.intersect(small) == small` is more concise, but I don't know if it would cause any floating-point precision issues.

lib/benchmarks/gallery_automator.dart
pennzht: [flutter/flutter — dev/benchmarks/macrobenchmarks/lib/src/web/gallery/gallery_recorder.dart:48
](https://github.com/flutter/flutter/pull/57576/files#diff-25d548fb6463b4b394e412dc939a415fR48)

```
  /// Whether we should continue recording.
  @override
  bool shouldContinue() => !_finished || profile.shouldContinue();

  GalleryAutomator _galleryAutomator;
  bool get _finished => _galleryAutomator?.finished ?? false;
```

lib/benchmarks/README.md
guidezpl: I see, the gallery is imported as a package, so everything needs to be in lib. Sounds good

lib/benchmarks/gallery_automator.dart
```diff
@@ -0,0 +1,197 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
willlarche, line 1: 2020. Check for all new files.

pubspec.yaml
```diff
@@ -40,14 +45,115 @@ dev_dependencies:
...
   assets:
-    - fonts/google_fonts/
-    - assets/demos/
-    - assets/icons/cupertino/
-    - assets/icons/material/
-    - assets/icons/reference/
-    - assets/icons/settings/
-    - assets/logo/
-    - assets/studies/
+    - packages/flutter_gallery_additional_assets/assets/icons/settings/settings_light.flr
```
guidezpl, line 48: You don't have to specify every size, just the "size-less" version

lib/benchmarks/gallery_automator.dart
pennzht: Updated new dart files to `2020`.

pubspec.yaml
pennzht: Done.

lib/pages/home.dart
```diff
@@ -155,9 +168,10 @@ class HomePage extends StatelessWidget {
                   FadeInImagePlaceholder(
                     image: Theme.of(context).colorScheme.brightness ==
                             Brightness.dark
-                        ? const AssetImage('assets/logo/flutter_logo.png')
-                        : const AssetImage(
-                            'assets/logo/flutter_logo_color.png'),
+                        ? const AssetImage('assets/logo/flutter_logo.png',
```
guidezpl, line 171: nit: trailing comma

lib/pages/splash.dart
```diff
@@ -182,7 +182,8 @@ class _SplashBackLayer extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     var effectAsset = 'splash_effects/splash_effect_$effect.gif';
-    final flutterLogo = Image.asset('assets/logo/flutter_logo.png');
+    final flutterLogo = Image.asset('assets/logo/flutter_logo.png',
+        package: 'flutter_gallery_assets');
```
guidezpl, line 186: nit: trailing comma

lib/data/demos.dart
```diff
@@ -79,6 +81,8 @@ class GalleryDemo {
     @required this.title,
     @required this.category,
     @required this.subtitle,
+    // This parameter is required for studies.
+    this.studyId,
```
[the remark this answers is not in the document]
pennzht: @perclasson 
Will it break anything if we add `slug` values to studies (Shrine, etc.)? It looks fine to me, but I'm unsure.

lib/pages/home.dart
```diff
@@ -45,44 +45,54 @@ class HomePage extends StatelessWidget {
     final carouselCards = <Widget>[
       _CarouselCard(
         demo: studyDemos['shrine'],
-        asset: const AssetImage('assets/studies/shrine_card.png'),
+        asset: const AssetImage('assets/studies/shrine_card.png',
+            package: 'flutter_gallery_assets'),
```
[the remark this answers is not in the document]
pennzht: Done.

lib/pages/splash.dart
pennzht: Done.

lib/pages/home.dart
pennzht: Done.

lib/data/demos.dart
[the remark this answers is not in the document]
perclasson: It should be fine to use `slug`. But it will allow us to use the studies via /demo/rally which is not intended.

We could simply update `slugToDemo` to only get the demos that are relevant:

```dart
/// Used to access code demos via slugs, used in the routes for the root Navigator.
Map<String, GalleryDemo> slugToDemo(BuildContext context) {
  final localizations = GalleryLocalizations.of(context);
  return LinkedHashMap<String, GalleryDemo>.fromIterable(
    materialDemos(localizations) +
    cupertinoDemos(localizations) +
    otherDemos(localizations),
    key: (dynamic demo) => demo.slug as String,
  );
}
```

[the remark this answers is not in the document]
pennzht: @guidezpl @perclasson 

Thank you for your message!

After some thought, I suggest keeping `slug` and `studyId` separate, in order to keep the code easier to understand.

PR #186, opened 2020-06-18
[Performance] Refactor SettingsListItem

### LINKED ISSUE #46: [Performance] Settings locale list initial load is slow (3)
### Acceptance Criteria
* Expand/collapse animation for settings should be smooth
* Look into different ListView constructor that recycles items

original issue: https://github.com/material-components/material-components-flutter-gallery/issues/346

guidezpl:
This PR refactor SettingsListItem for performance. 
- Only creates a RadioListTile for those visible items instead of all of them
- Limits the container height so that only 8 list items can be visible at once. This really only affects the behavior for the locale setting.

Closes #46

### Web startup performance
Before | After
------------ | -------------
before | after

Screenshot 2020-06-18 at 23 03 53
Screenshot 2020-06-18 at 21 35 34

### DIFF
--- a/lib/pages/settings_list_item.dart
+++ b/lib/pages/settings_list_item.dart
@@ -82,16 +82,9 @@ class SettingsListItem<T> extends StatefulWidget {
     @required this.onOptionChanged,
     @required this.onTapSetting,
     @required this.isExpanded,
-  })  : options = optionsMap.keys,
-        displayOptions = optionsMap.values,
-        super(key: key);
+  }) : super(key: key);
 
   final LinkedHashMap<T, DisplayOption> optionsMap;
-
-  /// For ease of use. Correspond to the keys and values of [optionsMap].
-  final Iterable<T> options;
-  final Iterable<DisplayOption> displayOptions;
-
   final String title;
   final T selectedOption;
   final ValueChanged<T> onOptionChanged;
@@ -115,6 +108,10 @@ class _SettingsListItemState<T> extends State<SettingsListItem<T>>
   Animation<EdgeInsetsGeometry> _headerPadding;
   Animation<EdgeInsetsGeometry> _childrenPadding;
   Animation<BorderRadius> _headerBorderRadius;
+
+  // For ease of use. Correspond to the keys and values of `widget.optionsMap`.
+  Iterable<T> _options;
+  Iterable<DisplayOption> _displayOptions;
 
   @override
   void initState() {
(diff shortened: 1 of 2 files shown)


lib/pages/settings_list_item.dart
```diff
@@ -76,16 +76,23 @@ class SlowMotionSetting extends StatelessWidget {
...
     @required this.onTapSetting,
     @required this.isExpanded,
-  }) : super(key: key);
+  })  : options = optionsMap.keys,
+        displayOptions = optionsMap.values,
+        super(key: key);
+
+  final LinkedHashMap<T, DisplayOption> optionsMap;
+
+  /// For ease of use. Correspond to the keys and values of [optionsMap].
```
rami-a, line 91: I feel like breaking the map into these 2 Iterables doesn't add much value. I suggest removing them.

If you think they should remain, then please make them private.

lib/pages/settings_list_item.dart
```diff
@@ -190,46 +197,11 @@ class _SettingsListItemState<T> extends State<SettingsListItem<T>>
...
-          dense: true,
-        ),
-      ),
-    );
-
     return AnimatedBuilder(
       animation: _controller.view,
       builder: _buildHeaderWithChildren,
       child: Container(
+        constraints: const BoxConstraints(maxHeight: 380),
```
rami-a, line 204: So is this the real crux of the fix? Just capping the height?

guidezpl: Yes, I believe most of the gains are gotten through this

lib/pages/settings_list_item.dart
guidezpl: Doing data structure operations felt kind of wrong in the build method, although that's purely based on feeling. I'll see to moving this to the _State class

guidezpl: Did a bit of reading on these things and because it's implemented as a hash map + doubly linked list, obtaining the keys/values is inexpensive.

lib/pages/settings_list_item.dart
clocksmith: since 380 / 8 = 47.5, and if this is to show exactly 8 items, should it maybe be something like 384? (maybe the height of the item is 48?)

guidezpl: Good catch, this makes it fit perfectly

PR #206, opened 2020-07-09
Add motion fade scale transition demo

### LINKED ISSUE #199: Add fade scale transition demos to reference section under motion
Add fade scale transition demo from animations package https://pub.dev/packages/animations

Renzo-Olivares:
This resolves #199 

new


integration_test/gallery_compile_test.dart
```diff
@@ -8,8 +8,8 @@ import 'package:test/test.dart';
 import 'package:path/path.dart' as path;
 
 // Benchmark size in kB.
-const int bundleSizeBenchmark = 4100;
-const int gzipBundleSizeBenchmark = 1004;
+const int bundleSizeBenchmark = 4124;
+const int gzipBundleSizeBenchmark = 1008;
```
perclasson, line 12: I think it's fine to increase this a bit more, so we're do not have to update it all the time. Perhaps to 4500 and 1200?

lib/demos/reference/motion_demo_fade_scale_transition.dart
```diff
@@ -0,0 +1,158 @@
...
+      reverseDuration: const Duration(milliseconds: 75),
+      vsync: this,
+    )..addStatusListener((status) {
+        setState(() {
+          // setState needs to be called to trigger a rebuild because
+          // the 'HIDE FAB'/'SHOW FAB' button needs to be updated based
+          // the latest value of [_controller.status].
+        });
+      });
+    super.initState();
```
perclasson, line 37: From the [documentation](https://api.flutter.dev/flutter/widgets/State/initState.html):
> If you override this, make sure your method starts with a call to super.initState().

lib/demos/reference/motion_demo_fade_scale_transition.dart
```diff
@@ -0,0 +1,158 @@
...
+                  onPressed: () {
+                    showModal<void>(
+                      context: context,
+                      builder: (context) {
+                        return _ExampleAlertDialog();
+                      },
+                    );
+                  },
+                  color: colorScheme.primary,
+                  textColor: colorScheme.onPrimary,
```
rami-a, line 104: I think the button should already be using these colors because of the theme, so you can probably remove these 2 lines here and below

lib/demos/reference/motion_demo_fade_scale_transition.dart
```diff
@@ -0,0 +1,158 @@
...
+    );
+  }
+}
+
+class _ExampleAlertDialog extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    final localizations = GalleryLocalizations.of(context);
+
+    return AlertDialog(
```
rami-a, line 138: In the gif it looks like the dialog doesn't follow the demo theme, I remember running into this when I did the dialog demo and resolved it by wrapping the widget with the theme, like here: https://github.com/flutter/gallery/blob/master/lib/demos/material/dialog_demo.dart#L41

lib/demos/reference/motion_demo_fade_scale_transition.dart
```diff
@@ -0,0 +1,158 @@
...
+        children: [
+          const Divider(height: 0),
+          Padding(
+            padding: const EdgeInsets.symmetric(vertical: 8),
+            child: Row(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                RaisedButton(
+                  onPressed: () {
+                    showModal<void>(
```
rami-a, line 96: I'm a bit confused about how this uses the Fade transition, isn't `showModal` just built in to flutter? Or is this from the animations package

Renzo-Olivares: Are you referring to 'showDialog'? Couldn't find 'showModal', but the one used here is from the animations package https://pub.dev/documentation/animations/latest/animations/showModal.html

integration_test/gallery_compile_test.dart
Renzo-Olivares: sounds good to me. I'll update it.

lib/demos/reference/motion_demo_fade_scale_transition.dart
Renzo-Olivares: Thanks I'll check this out.

lib/demos/reference/motion_demo_fade_scale_transition.dart
Renzo-Olivares: will do

lib/demos/reference/motion_demo_fade_scale_transition.dart
rami-a: Ah ok, my bad I was getting them confused!

lib/demos/reference/motion_demo_fade_scale_transition.dart
Renzo-Olivares: thanks! will change.

PR #213, opened 2020-07-15
Update to new android embedding v2

Renzo-Olivares:
Fixes warning when doing flutter run
`Your Flutter application is created using an older version of the Android embedding. It's being deprecated in favor of Android embedding v2. Follow the steps on https://flutter.dev/go/android-project-migration to migrate your project.`

Instructions to update outlined here: https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects


android/app/src/main/res/values/styles.xml
```diff
@@ -6,4 +6,13 @@
...
     </style>
     <color name="backgroundColor">#030303</color>
+    <!-- Theme applied to the Android Window as soon as the process has started.
+     This theme determines the color of the Android Window while your
+     Flutter UI initializes, as well as behind your Flutter UI while its
+     running.
+
+     This Theme is only used starting with V2 of Flutter's Android embedding. -->
+    <style name="NormalTheme" parent="@android:style/Theme.Black.NoTitleBar">
+        <item name="android:windowBackground">@android:color/white</item>
```
rami-a, line 16: Is this white the color of the splash screen when the android app launches? If so, can we make it black instead?

Renzo-Olivares: Sorry just saw this, the normal theme is the background that shows up briefly when changing the device orientation, briefly after the LaunchTheme, or during activity restoration. So far I haven't been able to capture it. The splash screen is set to black in the LaunchTheme, should we also set this one to black?

rami-a: yeah I think that makes sense

Issue #214, opened 2020-07-16
Add focused performance test for Crane ListView scrollling


clocksmith:
See: https://github.com/flutter/flutter/issues/61193#issuecomment-659069400

PR #216, opened 2020-07-17
Add performance transitions test for crane only

guidezpl:
Adds a test to perform the transitions test for Crane only. 

To run the transitions_perf test for all demos:
`flutter drive --profile --trace-startup -t test_driver/transitions_perf.dart -d <device>`
To run the transitions_perf test for just Crane:
`flutter drive --profile --trace-startup -t test_driver/transitions_perf.dart -d <device> --dart-define=onlyCrane=true`

Since --tags isn't yet supported for flutter driver tests and `skip` isn't designed for runtime determination, this is the simplest way to conditionally run a driver test from the command line.

Closes https://github.com/flutter/gallery/issues/214

### DIFF
--- a/test_driver/transitions_perf_test.dart
+++ b/test_driver/transitions_perf_test.dart
@@ -11,6 +11,11 @@ import 'package:test/test.dart' hide TypeMatcher, isInstanceOf;
 
 import 'isolates_workaround.dart';
 
+// To run this test for all demos:
+// flutter drive --profile --trace-startup -t test_driver/transitions_perf.dart -d <device>
+// To run this test for just Crane:
+// flutter drive --profile --trace-startup -t test_driver/transitions_perf.dart -d <device> --dart-define=onlyCrane=true
+
 // Demos for which timeline data will be collected using
 // FlutterDriver.traceAction().
 //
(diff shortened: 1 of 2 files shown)


test_driver/transitions_perf_test.dart
```diff
@@ -210,11 +222,31 @@ void main([List<String> args = const <String>[]]) {
...
+        },
+        streams: const <TimelineStream>[
+          TimelineStream.dart,
+          TimelineStream.embedder,
+        ],
+      );
+
+      final summary = TimelineSummary.summarize(timeline);
+      await summary.writeSummaryToFile('transitions-crane', pretty: true);
+    });
```
rami-a, line 241: The `all demos` test has a timeout, should this one?

guidezpl: I don't see why not, I'll add one

guidezpl:
For web, I never managed to get the `driver.scrollUntilVisible` command to work, and so the test will hang here https://github.com/flutter/gallery/blob/master/test_driver/transitions_perf_test.dart#L146 until it times out.

liyuqian:
When I tried to add this to Flutter's devicelab today, I found https://github.com/flutter/gallery/issues/227.

PR #219, opened 2020-07-17
[a11y-web] adding enabled to button semantics 

nturgut:
Web will add aria-disabled to these elements otherwise: https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/semantics/tappable.dart#L27

### DIFF
--- a/lib/pages/backdrop.dart
+++ b/lib/pages/backdrop.dart
@@ -298,6 +298,7 @@ class _SettingsIcon extends AnimatedWidget {
       child: Semantics(
         sortKey: const OrdinalSortKey(1),
         button: true,
+        enabled: true,
         label: _settingsSemanticLabel(isSettingsOpenNotifier.value, context),
         child: SizedBox(
           width: _settingsButtonWidth,
--- a/lib/pages/home.dart
+++ b/lib/pages/home.dart
@@ -1093,6 +1093,7 @@ class _StudyWrapperState extends State<StudyWrapper> {
                 sortKey: const OrdinalSortKey(0),
                 label: GalleryLocalizations.of(context).backToGallery,
                 button: true,
+                enabled: true,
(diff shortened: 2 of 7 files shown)


nturgut:
This is a quick fix for the gallery code before the july updates. However a better fix might be changing enabled to default true in this class: https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/basic.dart#L6634 

@goderbauer  shall I also send a PR to the framework repo?

goderbauer:
I don't think this is the right fix. The bug appears to be here:

https://github.com/flutter/engine/blob/74df174fa3d8d95fcfee654fec5cd37cc5d191a2/lib/web_ui/lib/src/engine/semantics/tappable.dart#L27-L29

Something should only be marked as disabled when the `hasEnabledState` is set and the `isEnabled` is not set. If `hasEnabledState` is not set, the element has no opinion of whether it is disabled or enabled.

goderbauer:
I would recommend reverting this change. It may have unintended a11y side-effects because (some of) the wrapped button already set this property: https://github.com/flutter/flutter/blob/5b9ce492f3da4a2b3a629ede34951bcac7d79a56/packages/flutter/lib/src/material/button.dart#L462

nturgut:
@goderbauer I see, so we were implementing it wrong on the engine side for a while. I'll revert this one.

goderbauer:
I took a closer look at where this PR is adding the `enabled: true` properties. It looks like it is doing this only in places where the semantics of the wrapped button are actually disabled via `excludeSemantics`. In these cases there's no chance of unintended side-effects (because the real button semantics are disabled) and the `enabled: true` is probably correct here.

Nevertheless, the web engine still has the bug that something is only disabled if `hasEnabledState` is set and `isEnabled` is not set.

PR #220, opened 2020-07-20
Revert "[a11y-web] adding enabled to button semantics "

nturgut:
Reverts flutter/gallery#219

We concluded that the issue is to be fixed in the web engine. Reverting this change.

### DIFF
--- a/lib/pages/backdrop.dart
+++ b/lib/pages/backdrop.dart
@@ -298,7 +298,6 @@ class _SettingsIcon extends AnimatedWidget {
       child: Semantics(
         sortKey: const OrdinalSortKey(1),
         button: true,
-        enabled: true,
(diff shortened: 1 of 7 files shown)


nturgut:
closing the PR: https://github.com/flutter/gallery/pull/219#issuecomment-661212238

goderbauer:
I actually take back my statement that this should be reverted, see https://github.com/flutter/gallery/pull/219#issuecomment-661212238.

Since it only adds `enabled: true` to places where the actual button semantics are disabled via `excludeSemantics`, adding `enabled: true` there is actually correct.

Nevertheless, the web engine still has the bug that it should only mark something as disabled when `hasEnabledState` is set and `isEnabled` is not set. 

Sorry about the confusion.

PR #221, opened 2020-07-20
InteractiveViewer in transformations demo

justinmc:
Flutter now includes a widget that does a lot of what the transformations demo does.  This PR swaps the widget into the PR and removes the dead code.

The demo appears a bit neglected and still has some bugs after swapping in InteractiveViewer.  I'll try to follow up on these in separate PRs.


csells:
lgtm

lib/demos/reference/transformations_demo.dart
```diff
@@ -17,29 +18,91 @@ class TransformationsDemo extends StatefulWidget {
...
-      board: _board,
+  void initState() {
+    super.initState();
+    _controllerReset = AnimationController(
+      vsync: this,
     );
+  }
 
+  @override
+  void didUpdateWidget(TransformationsDemo oldWidget) {
```
rami-a, line 98: Do you need this since all it's doing is calling super?

lib/demos/reference/transformations_demo.dart
```diff
@@ -3,10 +3,11 @@
 // found in the LICENSE file.
 
 import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
 import 'package:gallery/l10n/gallery_localizations.dart';
+
```
rami-a, line 8: Since you updated the demo, you'll need to update code segments so that the code viewer shows the correct code, https://github.com/flutter/gallery/blob/master/tool/codeviewer_cli/README.md

The CI will fail on this but its currently failing early because of #222

lib/demos/reference/transformations_demo.dart
justinmc: Nope, good catch.

lib/demos/reference/transformations_demo.dart
justinmc: Thanks, I was just going to ping you for help on getting CI green.  I updated the PR with master and updated the code segments, so hopefully it works now.

PR #224, opened 2020-07-21
Bottom nav animations revamp

Renzo-Olivares:
Switch animations in the bottom navigation bar demo to use the animations package. This reduces the added complexity that comes with setting up an animation controller, disposing of it, and manipulating a transition stack.

Before|After
---|---
bottom-nav-demo | bottom-nav-animations-revamp


lib/demos/material/bottom_navigation_demo.dart
```diff
@@ -152,60 +123,48 @@ class _NavigationIconView {
...
-              child: Padding(
-                padding: const EdgeInsets.all(16),
-                child: ClipRRect(
-                  borderRadius: BorderRadius.circular(8),
-                  child: Image.asset(
-                    'assets/demos/bottom_navigation_background.png',
-                    package: 'flutter_gallery_assets',
-                  ),
+
+  Stack destination(BuildContext context) {
```
rami-a, line 135: I'm not a big fan of effectively creating a build method inside of a non widget class. Could this be refactored such that this is a `StatelessWidget` that takes in a `BottomNavigationBarItem` in the constructor. Then rename/replace `_navigationViews` with a list of `BottomNavigationBarItem`s.

Renzo-Olivares: Will do.

PR #230, opened 2020-07-23
Bump version for motion demos

clocksmith:


pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: gallery
 description: A resource to help developers evaluate and use Flutter.
 repository: https://github.com/flutter/gallery
-version: 2.4.0+020400
+version: 2.4.1
```
guidezpl, line 4: Needs to be 2.4.1+020401

rami-a: yeah, this format was needed to make the play store happy

rami-a: Actually since this includes all those nice new motion demos, maybe 2.5.0 is more appropriate?

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: gallery
 description: A resource to help developers evaluate and use Flutter.
 repository: https://github.com/flutter/gallery
-version: 2.4.0+020400
+version: 2.5.0+020401
```
rami-a, line 4:
```suggestion
version: 2.5.0+020500
```

clocksmith: Done. How are these values after the + determined?

PR #231, opened 2020-07-28
Bump animations to 1.1.2

goderbauer:


rami-a:
Looks like golden tests is failing, this doesn't seem like it should affect that though

goderbauer:
@rami-a I updated the goldens, I fail to see a difference, though.

rami-a:
Yeah I also don't see a difference, but if it makes the check green I think that's ok

PR #242, opened 2020-07-30
Add shared z axis transition between login and home pages of Rally study

### LINKED ISSUE #237: Add z axis transition between Rally login and home pages
Currently the Rally study transitions from login to home using a sliding animation. This is a prime example for the z-axis transition animation, for transitioning between different levels of an application (login -> home).

Renzo-Olivares:
Closes #237 

Previously the default slide animation is used to transition between different screens in Rally. This change replaces that animation with the shared z axis transition from the animations package because navigating between different levels of an application is a prime use case of the material motion system's shared z axis transition.

Before|After
---|---
rally-old-transition | new-rally-z-axis-transition


lib/studies/rally/app.dart
```diff
@@ -22,13 +23,24 @@ class RallyApp extends StatelessWidget {
...
+  );
+
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
       title: 'Rally',
       debugShowCheckedModeBanner: false,
       theme: _buildRallyTheme().copyWith(
         platform: GalleryOptions.of(context).platform,
+        pageTransitionsTheme: PageTransitionsTheme(
```
rami-a, line 38: Would this also change the transition for going to other pages in Rally? I guess you already created a custom transition for going into an account page

PR #246, opened 2020-07-30
Fix and simplify transformation demo to ameliorate web issues

justinmc:
This PR makes the transformations demo much more presentable.

The transformation demo had several problems on web. This PR either fixes these problems or works around them when the app is running on web.

  1. Bad performance. 🟡   Simplified number of polygons being drawn. It's usable now but still choppy.  Issue: https://github.com/flutter/flutter/issues/60550
  1. Zooming caused weird visual artifacts instead of changing scale. 🟡  Worked around this by disabling scaling on web.
  1. The child wasn't centered depending on the window size. ✅  Fixed.
  1. Parts of the scene were not interactive. ✅  Fixed.


lib/demos/reference/transformations_demo.dart
```diff
@@ -79,7 +80,9 @@ class _TransformationsDemoState extends State<TransformationsDemo>
   }
 
   void _onTapUp(TapUpDetails details) {
-    final scenePoint = _transformationController.toScene(details.localPosition);
+    final renderBox = _targetKey.currentContext.findRenderObject() as RenderBox;
+    final offset = details.globalPosition - renderBox.localToGlobal(Offset.zero);
+    final scenePoint = _transformationController.toScene(offset);//details.localPosition);
```
rami-a, line 85: Nit: Remove commented out code

justinmc: Whoops thanks!

PR #262, opened 2020-08-17
Fix Cupertino Pickers theme

Vorkytaka:
Right now, when Cupertino dialog shows it use app theme, instead of parent widget theme.
But all examples shows with light theme.

If we use dark theme on device, then all pickers have very strange styles: it's use dark theme for background (cuz of `CupertinoColors.systemBackground.resolveFrom(context)` i think), but with black fonts.

The strangest behavior in Timer example (last screenshot).

With this fix pickers will use parent widget theme.

### Before:
Screenshot_20200817-193927

### After:
Screenshot_20200817-193327

### Timer without fix:
Screenshot_20200817-200317


Vorkytaka:
@googlebot I signed it!

lib/demos/cupertino/cupertino_picker_demo.dart
```diff
@@ -27,24 +27,36 @@ class _CupertinoPickerDemoState extends State<CupertinoPickerDemo> {
   // Value that is shown in the date picker in dateAndTime mode.
   DateTime dateTime = DateTime.now();
 
+  void _showDemoPicker(
+      {@required BuildContext context, @required Widget child}) {
```
rami-a, line 31: Please add a trailing comma after child and reformat.

lib/demos/cupertino/cupertino_picker_demo.dart
```diff
@@ -27,24 +27,36 @@ class _CupertinoPickerDemoState extends State<CupertinoPickerDemo> {
   // Value that is shown in the date picker in dateAndTime mode.
```
rami-a, line 27: Since you changed the code in a demo file, please run:

```
flutter pub run grinder verify-code-segments
```

to update the code segments, as described here: https://github.com/flutter/gallery/#generating-localized-strings-and-highlighted-code-segments

rami-a:
Thanks for the fix @Vorkytaka, just a couple of small comments and then this should be good to merge

PR #265, opened 2020-08-18
update to latest path_provider_interface to unblock framework null-safety

jonahwilliams:
Need to pull in the latest path_provider_interface, updates the lockfile


jonahwilliams:
Thanks!

PR #272, opened 2020-08-24
Adaptive breakpoint uses operator logic

JoseAlba:


lib/layout/adaptive.dart
```diff
@@ -9,9 +9,7 @@ import 'package:adaptive_breakpoints/adaptive_breakpoints.dart';
 /// build adaptive and responsive layouts.
 bool isDisplayDesktop(BuildContext context) {
```
rami-a, line 10: You could make this even shorter
```suggestion
bool isDisplayDesktop(BuildContext context) => getWindowType(context) >= AdaptiveWindowType.m;
```

JoseAlba: Done

PR #282, opened 2020-09-01
Use Synthetic Packages for localizations

shihaohong:
With synthetic package support in the gen_l10n tool (https://github.com/flutter/flutter/pull/62395), we no longer need to check in generated localizations .dart files into any Flutter project. Here is the laundry list of changes involved in the PR:

1. This PR introduces an `l10n.yaml` file to set up localizations generation when `flutter test`, `flutter pub get`, `flutter analyze` are invoked. It also generates localizations on `flutter run`, hot reload and hot restart so that direct modifications to the arb files can be made and observed while in development.
2. Removal of all generated `.dart` localizations files from source control, since it's no longer needed.
3. Updated all imports to point to the synthetic package location.
4. Slight update to the `grind.dart` tool that formats the generated code. I left the rest of the `grind.dart` command in there because it seems like we running `flutter pub get` in `${FLUTTER}/dev/tools` is unfortunately still a necessity. My hope is to get rid of this requirement altogether, but I plan on doing this in a subsequent PR.
5. Edit: Add `flutter update-packages` as a step in the CI temporarily since without it (issue filed https://github.com/flutter/flutter/issues/65023), `flutter pub get` does not work as expected since the gen_l10n tool's dependencies are not always in sync. This is particularly observed when switching between versions of flutter (ie. upgrading from stable to master branch, or vice versa)


shihaohong:
The tests are failing because the `gen_l10n` tool's dependencies aren't being properly created unless you run `flutter update-packages`, which is similar to the issue seen in https://github.com/flutter/samples/issues/331. I added a step to the tests where `flutter update-packages` is run before `flutter pub get`, which now generates synthetic packages.

My next goal is to update the `gen_l10n` tool itself so that this no longer becomes a problem.

tool/grind.dart
```diff
@@ -44,7 +44,6 @@ Future<void> generateLocalizations() async {
     '--preferred-supported-locales=["en"]',
     '--use-deferred-loading',
```
perclasson, line 45: Are these arguments still needed now with the l10n.yaml file? Or is it simply enough to just run gen_l10n.dart?

perclasson (approved):
LGTM, this is great!

tool/grind.dart
shihaohong: The command itself shouldn't be necessary at all anymore, since localizations generation is baked into Flutter run, restart, reload, pub get, analyze and test. 

The only reason I've left it in for now is because there's an issue with the gen_l10n tool's having the incorrect dependencies when switching between flutter versions that I still need to fix. Running the grinder command will always work, whereas the synthetic package generation may fail unless `flutter update-packages` is run.

clocksmith:
I am having some trouble running the Gallery from a fresh clone. 

`flutter upgrade` is at 
```
Flutter is already up to date on channel master
Flutter 1.22.0-10.0.pre.39 • channel master • https://github.com/flutter/flutter.git
Framework • revision a8281e31af (23 hours ago) • 2020-09-01 10:57:59 -0700
Engine • revision 165abef0a2
Tools • Dart 2.10.0 (build 2.10.0-77.0.dev)
```

and after running `flutter update-packages`, GalleryLocalizations is unable to be resolved:

Screen Shot 2020-09-02 at 12 23 01 PM

clocksmith:
`flutter clean`, then `flutter pub get` recreated `.dart_tool` with the `flutter_gen` inside, so it works now. Thank you @Renzo-Olivares !

xster:
Is flutter gallery expected to run on flutter stable too?

I'm getting 

```
Error detected in pubspec.yaml:
Unexpected child "generate" found under "flutter".
Please correct the pubspec.yaml file at /Users/xster/development/flutter-open/gallery/pubspec.yaml
``` 

if I run on stable.

rami-a:
Nope, gallery targets master https://github.com/flutter/gallery#running-flutter-gallery-on-flutters-master-channel so its expected that things won't always work on stable.

xster:
Understood, thanks for clarifying.

xster:
I've been getting this LUCI test issue when I sync flutter/flutter past this commit. 

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket.appspot.com/8870127907410516896/+/steps/run_test.dart_for_hostonly_devicelab_tests_shard_and_subshard_3_last/0/stdout

Do we need to modify dev/bots/test.dart to unblock flutter/flutter?

627171314 (commented):
image

guidezpl:
@627171314  lease follow the steps at https://github.com/flutter/gallery/pull/282#issuecomment-685867513, if that doesn't work, please file a separate issue.

guidezpl:
@627171314  please follow the steps at https://github.com/flutter/gallery/pull/282#issuecomment-685867513, if that doesn't work, please file a separate issue.

PR #268, opened 2020-08-21
Increase Crane performance test timeout

liyuqian:
See https://github.com/flutter/flutter/issues/64328#issuecomment-678412127

We'll need to also roll the new gallery version in https://github.com/flutter/flutter/blob/master/dev/devicelab/lib/versions/gallery.dart to have Flutter use this.

PR #283, opened 2020-09-02
Remove test timeout

liyuqian:
See https://github.com/flutter/gallery/pull/268#issuecomment-680292377


guidezpl (approved):
Do we want to keep the 2 scroll timeouts?

liyuqian:
I think we can keep scroll timeout now until we see problems from them. Previously, if a scroll timed out, it's usually a genuine failure (e.g., connection failed, element not found), so it seems to be good to let that case fail in 10 seconds instead of 60 minutes.

PR #287, opened 2020-09-03
[Reply] Refactor BottomAppBar to AnimatedBottomAppBar and fix email tappable bug

Renzo-Olivares:
This change brings in changes from the reply codelab.
* Refactor `bottomNavigationBar` property of `MobileNav` `Scaffold` to a separate class, `_AnimatedBottomAppBar` to improve readability.
* Wrap `BottomAppBar` in a `Container` instead of `SizedBox` to fix issue where emails could still be tapped even though they were not visible under the `BottomAppBar`


lib/studies/reply/adaptive_nav.dart
```diff
@@ -791,6 +732,99 @@ class _MobileNavState extends State<_MobileNav> with TickerProviderStateMixin {
...
+  final List<_Destination> navigationDestinations;
+  final int selectedIndex;
+  final VoidCallback toggleBottomDrawerVisibility;
+
+  @override
+  Widget build(BuildContext context) {
+    return Selector<EmailStore, bool>(
+      selector: (context, emailStore) => emailStore.onMailView,
+      builder: (context, onMailView, child) {
+        bottomAppBarController.forward();
```
guidezpl, line 761: I don't think bottomAppBarController needs to be passed down. You can do this in the parent's lifecycle functions

guidezpl (approved):
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
                        
```

lib/studies/reply/adaptive_nav.dart
Renzo-Olivares: I tried putting it in `didUpdateWidget` and `didChangeDependencies`, but those aren't called when we transition to the `MailViewPage`.

PR #292, opened 2020-09-10
[Motion] Fix exit page not transitioning out on shared z axis transition demo

Renzo-Olivares:
The home route was not wrapped in any transitions, so when we transition to the search route it does not animate and remains static. This change wraps our home route in a `PageRouteBuilder` and gives it the `SharedAxisTransition` that we want.

https://github.com/Renzo-Olivares/gallery/pull/47 similar to issue that was going on here.


rami-a (approved):
LGTM

clocksmith (approved):

PR #299, opened 2020-09-15
Updates to the Reply Navigation Rail animation

clocksmith:
This change makes a few updates to the Nav Rail extended animation:

- Fade in trailing widget
- Immediately start the animation on press
- Animate the width of the leading (and FAB) and trailing widgets
  * The FAB still has to "jump" from a non-extended FAB (56 height) to an extended FAB (44 height) because there is currently no easy way to lerp between these 2 shapes
- All nav rail children animations are now driven by the extended animation of the nav rail
  * An issue in flutter/flutter to correct the curve and duration: https://github.com/flutter/flutter/issues/65121

The destinations also no longer jump by a combination of https://github.com/flutter/flutter/pull/65659 and correcting the leading animation.

Before regular speed:
current_1x
After regular speed:
new_1x

Before 5x slow speed:
current_5x
After 5x slow speed:
new_5x


lib/studies/reply/adaptive_nav.dart
```diff
@@ -240,46 +203,40 @@ class _DesktopNavState extends State<_DesktopNav>
...
                                   ),
+                                  label: Text(destination.name),
                                 ),
-                              ),
-                              label: Text(destination.name),
+                            ],
+                            extended: _isExtended.value,
+                            labelType: NavigationRailLabelType.none,
+                            leading: _NavigationRailHeader(
+                              extended: _isExtended,
```
Renzo-Olivares, line 228: Nit: can `_NavigationRailHeader` fit on one line?

lib/studies/reply/adaptive_nav.dart
```diff
@@ -415,68 +370,87 @@ class _NavigationRailFolderSection extends StatelessWidget {
...
                       ),
-                      const SizedBox(height: 72),
-                    ],
-                  ),
-                ],
+                    ),
+                    const SizedBox(height: 8),
+                    for (var folder in folders.keys)
+                      InkWell(
+                        borderRadius: const BorderRadius.all(
```
Renzo-Olivares, line 416: nit: can `borderRadius` fit on one line?

Renzo-Olivares (approved):
This is Awesome! 

LGTM

lib/studies/reply/adaptive_nav.dart
clocksmith: It does not fit

lib/studies/reply/adaptive_nav.dart
clocksmith: It does not fit

rami-a (approved):
LGTM

PR #310, opened 2020-09-16
[Reply] Get bottom drawer background and mailbox title to be in sync with drawer animation

Renzo-Olivares:
This change brings in the following from the Reply motion codelab https://github.com/material-components/material-components-flutter-motion-codelab/pull/16 .

This change has the opacity of the bottom drawer background and mailbox title animate based on the drawer animation. It also adds a `FadeThroughAnimation` for when the mailbox title disappears when going from mailbox to mailview.

reply-mailbox-fade-through


lib/studies/reply/adaptive_nav.dart
```diff
@@ -726,6 +724,10 @@ class _AnimatedBottomAppBar extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    var fadeOut = Tween<double>(begin: 1, end: -1).animate(
```
rami-a, line 727: Why does this go to -1?

rami-a (approved):

lib/studies/reply/adaptive_nav.dart
Renzo-Olivares: I do this so the animation moves backwards instead of forwards. FadeTransition doesn't have a way to specify a reverse, so our end < begin. 

Also it's -1 because we want the mailbox title to be invisible by the time the drawer expands half way. So by setting this to -1 at the halfway point the opacity will be 0.

clocksmith (approved):

PR #323, opened 2020-09-22
Awaits the TickerFuture returned by AnimationController.reverse()

zanderso:
Possibly not caught before as it returns an implementer of Future rather than a straight-up Future.


rami-a (approved):
LGTM

PR #325, opened 2020-09-23
Update dependencies and platform directories

gspencergoog:
## Description

This updates the url_launcher to `^5.6.0` so that it includes Windows, and adds the missing platform directories for web, windows, and linux.

# Related Issues

- Fixes https://github.com/flutter/flutter/issues/65862


android/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml
```diff
@@ -1,5 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
rami-a, line 1: Why are there all of these Android changes? Also this looks like it is removing our Android adaptive icon. I don't think we wanna do that.

ios/Podfile
```diff
@@ -1,41 +0,0 @@
-# Uncomment this line to define a global platform for your project
```
rami-a, line 1: Same questions as above, why does this change delete a bunch of iOS related files?

macos/Podfile
```diff
@@ -1,82 +0,0 @@
-platform :osx, '10.11'
```
rami-a, line 1: Should we be deleting these macOS files?

macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -28,7 +28,6 @@
 		33CC11132044BFA00003C045 /* MainFlutterWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CC11122044BFA00003C045 /* MainFlutterWindow.swift */; };
```
rami-a, line 28: It looks like you deleted the gallery macOS app icon and replaced it with the default flutter icon. Github wouldn't let me comment on the image so I commented here. I don't think we want to do that.

stuartmorgan:
>  adds the missing platform directories for web, windows, and linux

There's already a web directory on trunk. I would suggest reverting all the ios/, android/, and web/ changes in this PR so that it's scoped specifically to adding Linux and Windows support, and doesn't have unintended side-effects on other platforms.

web/index.html
```diff
@@ -12,15 +12,9 @@
   <link rel="apple-touch-icon" href="icons/Icon-192.png">
 
   <!-- Favicon -->
-  <link rel="icon" type="image/png" href="favicon-16x16.png" sizes="16x16"/>
```
rami-a, line 15: This change also removed our custom gallery favicon

rami-a (commented):
I think adding the windows/linux directories makes sense. But it looks like this change is also modifying android/iOS/macOS/web in ways we don't necessarily want, such as removing the custom app icons

stuartmorgan:
Is there a definitive source for the Gallery icon? At some point (probably not in this PR) the default `.ico` in the Windows folder should be replaced with one for Gallery.

macos/Runner.xcodeproj/project.pbxproj
gspencergoog: Yeah, I already reverted that. Thanks for pointing it out.

rami-a:
> Is there a definitive source for the Gallery icon? At some point (probably not in this PR) the default `.ico` in the Windows folder should be replaced with one for Gallery.

Yes, I found the original sketch file that the designers created and shared it with the flutter team group. Also the highest res iOS asset could be used as a source of truth as it is 1024x1024

guidezpl:
> Is there a definitive source for the Gallery icon? At some point (probably not in this PR) the default `.ico` in the Windows folder should be replaced with one for Gallery.

A few of us have sketch files, but this one should suffice as a source https://github.com/flutter/gallery/blob/master/web/icons/Icon-512.png

gspencergoog:
I actually already reverted most of the macos/ios/android changes. Sorry, I was a little hasty in posting the PR, and I hadn't fully checked it on all the platforms yet. I was mainly creating it to see if the checks would pass, and I didn't think anyone would look at it at 6am PST. :-)

I did want to update the versions (kotlin, gradle, min OS version, etc.) in the macos/ios/android directories, but I can do that in another PR.

gspencergoog:
I did update the default windows icon though: we already had an .ico file for it in the tree.

stuartmorgan:
> I did update the default windows icon though: we already had an .ico file for it in the tree.

It looks like it's a web favicon that only has 16x16, unless GitHub's UI isn't giving me the right bits. The default Windows icon has 256, 48, 32, and 16, which based on some research at the time seemed to be the right set to cover for Win32 apps.

gspencergoog:
OK, I updated it for those sizes.

rami-a (approved):
LGTM

gspencergoog:
Actually, I'm going to make a better version: the small sizes in that .ico are not great.

Issue #346, opened 2020-10-13
Rally login button is wrong height


perclasson:
It does not have the correct height, should be about 40 dp height, something like this:
<img width="508" alt="Screen Shot 2020-10-13 at 16 56 14" src="https://user-images.githubusercontent.com/1770678/95879331-898db680-0d76-11eb-840b-64a85f949052.png">

ayaankhan98:
Basically the text `LOGIN` within the button is at wrong height, isn't?
![re](https://user-images.githubusercontent.com/43348292/95935550-1d03cd80-0df1-11eb-8dfc-a00a497a9914.png)

perclasson:
@ayaankhan98 Feel free to work in it :) that's very much appreciated.

Yes it looks as if the text is at the wrong height, but also that the vertical padding around it is too small. With the text, the button should be about 40px in height, right now it looks smaller to me.

ayaankhan98:
@perclasson i am facing some issue in local setup
<details>


<summary>Error logs</summary>

"af": 33 untranslated message(s).
"am": 33 untranslated message(s).
"ar": 33 untranslated message(s).
"ar_EG": 33 untranslated message(s).
"ar_JO": 33 untranslated message(s).
"ar_MA": 33 untranslated message(s).
"ar_SA": 33 untranslated message(s).
"as": 33 untranslated message(s).
"az": 33 untranslated message(s).
"be": 33 untranslated message(s).
"bg": 33 untranslated message(s).
"bn": 33 untranslated message(s).
"bs": 33 untranslated message(s).
"ca": 33 untranslated message(s).
"cs": 33 untranslated message(s).
"da": 33 untranslated message(s).
"de": 33 untranslated message(s).
"de_AT": 33 untranslated message(s).
"de_CH": 33 untranslated message(s).
"el": 33 untranslated message(s).
"en_AU": 33 untranslated message(s).
"en_CA": 33 untranslated message(s).
"en_GB": 33 untranslated message(s).
"en_IE": 33 untranslated message(s).
"en_IN": 33 untranslated message(s).
"en_NZ": 33 untranslated message(s).
"en_SG": 33 untranslated message(s).
"en_ZA": 33 untranslated message(s).
"es": 33 untranslated message(s).
"es_419": 33 untranslated message(s).
"es_AR": 33 untranslated message(s).
"es_BO": 33 untranslated message(s).
"es_CL": 33 untranslated message(s).
"es_CO": 33 untranslated message(s).
[...]

PR #348, opened 2020-10-15
fix: login button text margin

ayaankhan98:
Issue #346 
is it fine or any other modification?
button

### DIFF
--- a/lib/studies/rally/login.dart
+++ b/lib/studies/rally/login.dart
@@ -397,7 +397,7 @@ class _FilledButton extends StatelessWidget {
       style: TextButton.styleFrom(
         backgroundColor: RallyColors.buttonColor,
         primary: Colors.black,
-        padding: const EdgeInsets.symmetric(vertical: 14, horizontal: 24),
+        padding: const EdgeInsets.symmetric(vertical: 10, horizontal: 24),
         shape: RoundedRectangleBorder(
           borderRadius: BorderRadius.circular(12),
         ),
@@ -407,13 +407,7 @@ class _FilledButton extends StatelessWidget {
         children: [
           const Icon(Icons.lock),
           const SizedBox(width: 6),
-          Container(
-            margin: const EdgeInsets.only(top: 4.0),
-            child: SizedBox(
-              height: 14,
-              child: Text(text),
-            ),
-          ),
+          Text(text),
         ],
       ),
     );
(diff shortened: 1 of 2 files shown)


lib/studies/rally/login.dart
```diff
@@ -397,7 +397,7 @@ class _FilledButton extends StatelessWidget {
       style: TextButton.styleFrom(
         backgroundColor: RallyColors.buttonColor,
         primary: Colors.black,
-        padding: const EdgeInsets.symmetric(vertical: 10, horizontal: 24),
+        padding: const EdgeInsets.symmetric(vertical: 14, horizontal: 24),
```
perclasson, line 400: As mentioned in the ticket, the smaller Rally login button on desktop is probably due to [`VisualDensity`](https://api.flutter.dev/flutter/material/VisualDensity-class.html) being set to `adaptivePlatformDensity` and thus `compact` for Rally in desktop.

With your change the Button would become bigger on Android and iPhone phones. We will want to set the `visualDensity` to [`VisualDensity.standard`](https://api.flutter.dev/flutter/material/VisualDensity/standard-constant.html) instead.

To do this update the `ThemeData` for Rally: https://github.com/flutter/gallery/blob/2900cb93aeed58c953bda570774894ec7d30b08c/lib/studies/rally/app.dart#L58

ayaankhan98:
@perclasson after setting the `visualDensity` to `standard` this is what we have

#### Linux
linux

#### Web (Chrome)
chrome

everything looks good except the `Text`, `Text` should be at the same base level as that of `lock icon`. To fix this should i set the margin again in `lib/studies/rally/login.dart`?

perclasson:
> @perclasson after setting the `visualDensity` to `standard` this is what we have
> 
> 
> 
> #### Linux
> 
> linux
> 
> 
> 
> #### Web (Chrome)
> 
> chrome
> 
> 
> 
> everything looks good except the `Text`, `Text` should be at the same base level as that of `lock icon`. To fix this should i set the margin again in `lib/studies/rally/login.dart`?
> 
> 

This looks good. I think we do not need to modify the base level of the icon and text👍🏻

ayaankhan98:
okay no problem :)

perclasson (approved):
LGTM, thanks for the fix and dealing with the back and forth!

Issue #357, opened 2020-10-22
Use different l10n.yaml when developing vs deploying the web version


perclasson:
Currently we get warnings when working on the Gallery, that deferred loading causes hot reload to not work. We actually only want deferred loading to be used in the release version of the Gallery on web.

Figure out if it's possible to have a separate `l10n.yaml` for the web deploy version and for when developing on the Gallery.

PR #362, opened 2020-10-26
Only use deferred loading when building the web app

perclasson:
Currently we get warnings when working on the Gallery, that deferred loading causes hot reload to not work. We actually only want deferred loading to be used in the release version of the Gallery on web.

Closes https://github.com/flutter/gallery/issues/357


rami-a (commented):
LGTM

rami-a (approved):
LGTM

clocksmith (approved):

PR #366, opened 2020-11-01
Dynamic is not subtype of Double.

Abhishek01039:
we need to typecast of dynamic to double because `Color(0xFFF0EA).withOpacity` needs double.


Abhishek01039:
@shihaohong 
can you please review this PR?
Thanks

shihaohong:
It seems to be failing a Dart format check. Make sure to run `flutter format .` in the gallery since they follow a different style guide from the `flutter/flutter` repo.

Also, can you describe how this is causing an issue? It's not immediately clear to me why this update is necessary.

Abhishek01039:
Hey, @shihaohong Thanks for reviewing my PR.
oh my bad. I don't know why the compiler warned the error to me at that time but it is not causing an error.
so sorry for that
I'm closing this PR.
Thanks.

PR #368, opened 2020-11-05
[Reply] Partially State Restorable Pages

shihaohong:
This makes most of the pages in the Reply app state restorable. 

This PR does not yet include restoration of scrollable offset location, which I will complete in a separate PR.

State Restoration for Flutter is currently only enabled for Android (adding support for iOS is tracked in flutter/flutter#62915). In order to test the state restoration aspect of this app follow these steps:

1. Enable "Don't keep activities" in the developer settings of your Android device.
2. Run the app on the Android device via flutter run.
3. Navigate into the Reply app. Open up the search page or navigate to a different mailbox.
4. Switch to the home screen or another app via the app switcher. This kills the gallery app.
5. Switch back the gallery app via the app switcher. This will restart the gallery app.
6. Observe that the app is still on the original page.

This change required modifications to the email store logic and mail view logic since the state was so tightly coupled that it was difficult to make it state restorable without decoupling it. As nice side effects, the change removes duplicated state in the code,  improves starring/unstarring mail handling. For example, before this change, starred emails would not appear in the other mailboxes (inbox, sent, draft), because it had its own unique category. Now, starred emails will still appear on those aforementioned pages.

## Outstanding issue

State restoration is blocked on the mail view and compose pages due to the use of OpenContainer, which does not support the restorable Navigator APIs: https://github.com/flutter/flutter/issues/69924. This is causing a crash when using an Android device when:
1) On the Reply app, open a specific email. (ie. Inbox -> tap on any email by tapping the preview card)
2) Switch to the home screen or another app via the app switcher. This kills the gallery app.
3) Navigate around the different mailboxes, which will eventually trigger the crash:

```

════════ Exception caught by widgets library ═══════════════════════════════════
The following assertion was thrown building Navigator-[LabeledGlobalKey<NavigatorState>#37c7a](dirty, dependencies: [UnmanagedRestorationScope, HeroControllerScope, _EffectiveTickerMode], state: NavigatorState#2a1b0(tickers: tracking 1 ticker)):
'package:flutter/src/widgets/navigator.dart': Failed assertion: line 5104 pos 12: 'ild(BuildContext co': is not true.

Either the assertion indicates an error in the framework itself, or we should provide substantially more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=2_bug.md

The relevant error-causing widget was
Navigator-[LabeledGlobalKey<NavigatorState>#37c7a]
lib/…/reply/adaptive_nav.dart:1062
When the exception was thrown, this was the stack
#2      NavigatorState.build
package:flutter/…/widgets/navigator.dart:5104
#3      StatefulElement.build
package:flutter/…/widgets/framework.dart:4792
#4      ComponentElement.performRebuild
package:flutter/…/widgets/framework.dart:4675
#5      StatefulElement.performRebuild
package:flutter/…/widgets/framework.dart:4847
#6      Element.rebuild
package:flutter/…/widgets/framework.dart:4369
...


```


lib/studies/reply/model/email_model.dart
```diff
@@ -17,3 +19,46 @@ class Email {
...
+          recipients: recipients,
+          containsPictures: containsPictures,
+        );
+
+  InboxType inboxType;
+}
+
+// A set of the different mailbox pages that the Reply app
+// contains. An enum would be preferable, but only primitives
+// values are state restorable.
```
goderbauer, line 50: I agree, an enum would be better: Can you use an enum and store the index of the enum value (available via the `.index` getter on an enum value) in the restoration state? To restore the original enum, you can use that index as an index into the `values` constant of the enum.

lib/studies/reply/mail_card_preview.dart
```diff
@@ -26,16 +28,15 @@ class MailPreviewCard extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
 
-    final currentEmailStarred = Provider.of<EmailStore>(
-      context,
-      listen: false,
-    ).isEmailStarred(email);
-
+    // TODO(shihaohong): State restoration of mail view page is
```
goderbauer, line 38: Oh, did you by any chance take a look at OpenContainer to see what it would take to make that restorable?

goderbauer (commented):
This is looking great!

Can you also add some tests to ensure that we don't break state restoration for this app in the future? You can probably take some inspirations from the tests I added here: https://github.com/flutter/samples/pull/433

goderbauer:
I saw you also updated https://github.com/flutter/flutter/issues/62916 with the additional work we need to do! Thanks! Can you also link these migration PRs against the issue so we have them all in one place? Thanks!

goderbauer:
We also need to investigate the crash you are reporting in the PR description. Even if OpenContainer doesn't support state restoration, it shouldn't crash...

lib/studies/reply/mail_card_preview.dart
shihaohong: Discussion continued in https://github.com/flutter/flutter/issues/69924.

lib/studies/reply/app.dart
```diff
@@ -44,38 +66,86 @@ class ReplyApp extends StatelessWidget {
...
+    appState.selectedMailboxPage = MailboxPageType.values[mailboxPageIndex];
+
+    final starredEmailIdsList = appData['starredEmailIds'] as List<dynamic>;
+    appState.starredEmailIds = {
+      ...starredEmailIdsList.map<int>((dynamic id) => id as int),
+    };
+    final trashEmailIdsList = appData['trashEmailIds'] as List<dynamic>;
+    appState.trashEmailIds = {
+      ...trashEmailIdsList.map<int>((dynamic id) => id as int),
+    };
```
perclasson, line 131: nit: I think this can be just:

```dart
    appState.starredEmailIds = (appData['starredEmailIds'] as List<int>).toSet();
    appState.trashEmailIds = (appData['trashEmailIds'] as List<int>).toSet();
```

perclasson (approved):
LGTM!

lib/studies/reply/app.dart
shihaohong: When I tried this, I get a _CastError:
```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following _CastError was thrown building MediaQuery(MediaQueryData(size: Size(800.0, 600.0),
devicePixelRatio: 3.0, textScaleFactor: 1.0, platformBrightness: Brightness.light, padding:
EdgeInsets.zero, viewPadding: EdgeInsets.zero, viewInsets: EdgeInsets.zero, alwaysUse24HourFormat:
false, accessibleNavigation: false, highContrast: false, disableAnimations: false, invertColors:
false, boldText: false, navigationMode: traditional)):
type 'List<dynamic>' is not a subtype of type 'List<int>' in type cast

The relevant error-causing widget was:
  MediaQuery file:///Users/shihaohong/dev/apps/gallery/lib/data/gallery_options.dart:180:21

When the exception was thrown, this was the stack:
#0      _RestorableEmailState.fromPrimitives (package:gallery/studies/reply/app.dart:124:60)
#1      RestorationMixin.registerForRestoration (package:flutter/src/widgets/restoration.dart:845:20)
#2      _ReplyAppState.restoreState (package:gallery/studies/reply/app.dart:50:5)
```

I find type casting from dynamic to be pretty confusing and haven't found a sane way to determine if I can cast something a certain way other than trial and error. Any more ideas for this one?

perclasson: Oh strange, I don't have an idea. Let's drop this then, it does not seem worth to simplify it.

Issue #376, opened 2020-11-17
Snackbar not working as expected


Alabhya268:
**Describe the bug**
Snackbar not working as expected .
Whenever snackbar shows up , Its shows 2 snackbar both of different looks and size,
one uses whole app and other uses only demo section .

To switch channel run:
```bash
flutter channel master
flutter upgrade
```

**Flutter doctor**
Copy and paste the output of `flutter doctor -v` here

Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel master, 1.24.0-8.0.pre.266, on Microsoft Windows [Version 10.0.19042.630], locale en-US)
[√] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[√] Visual Studio - develop for Windows (Visual Studio Community 2019 16.8.1)
[√] Android Studio (version 4.1.0)
[√] VS Code (version 1.51.1)
[√] Connected device (1 available)

• No issues found!
```
```

**To Reproduce**
1. Go to 'Material demo section'
2. Click on 'snackbar demo'
3. Click on 'SHOW A SNACKBAR' button

**Expected behavior**
Only one snackbar should show up and it should only use demo area .

**Screenshots**
![snackbar1](https://user-images.githubusercontent.com/57143358/99376995-abd1a180-28eb-11eb-9811-10f05d8e9409.png)
![snackbar2](https://user-images.githubusercontent.com/57143358/99377004-ae33fb80-28eb-11eb-8fe5-928fd31ad00f.png)

**Desktop (please complete the following information):**
 - OS: Windows, Android

**Smartphone (please complete the following information):**
 - Device: Oneplus 7 pro
 - OS: Android 10 

**Additional context**
 Issue caused after merging this PR #343 .

rami-a:
hmm interesting, I guess this is an unintended side effect of the new snackbar messenger system. Perhaps because this screen contains two scaffolds, it is doing this.

@Piinks do you happen to know if there is away around this?

Piinks:
Hey @Alabhya268 thanks for reporting this! Yes, this is an unintended consequence of the ScaffoldMessenger. We are adding some error handling to prevent this in https://github.com/flutter/flutter/pull/70227 and updating the migration guide in https://github.com/flutter/website/pull/4977

[...]

PR #377, opened 2020-11-18
Scope SnackBars for nested Scaffolds in demo

Piinks:
Fixes https://github.com/flutter/gallery/issues/376


rami-a (approved):
Thanks for the fix Kate!

Though I believe there are a few more of these demos that also show a snackbar, they'd probably have to be updated too (dialog_demo, menu_demo, etc.)

Piinks:
Ah! thank you! I will run through and check all of them. :)

Piinks:
Found a much cleaner fix across the board. :) All SnackBars are properly scoped now.

PR #378, opened 2020-11-18
Restorable Material Code Samples

shihaohong:
This introduces state restoration to:
#### Demo
1. Go on the home page
2. Scroll down and under "Categories", tap "MATERIAL", "CUPERTINO", or "STYLES & OTHER". 
3. Kill the Android app and reopen, notice that the scroll location and the expansion states are preserved

#### Demo
1. Go on the home page
2. Scroll down and under "Categories", tap "MATERIAL", "CUPERTINO", or "STYLES & OTHER". 
3. Tap on any of the demos.
4. Kill the app, notice that the the page is preserved.

### The state of the some Material demo pages
1. Banner
2. Bottom app bar
3. Bottom navigation

#### Demo
1. Go on the home page
2. Scroll down and under "Categories", tap "MATERIAL", "CUPERTINO", or "STYLES & OTHER". 
3. Tap on any of the demos.
4. Tap any of the action icons in the AppBar (except the flutter API doc one).
5. Kill the app, notice that the page's state is preserved (Code sample is shown, or description of the widget is shown).


lib/pages/category_list_item.dart
```diff
@@ -70,9 +77,7 @@ class _CategoryListItemState extends State<CategoryListItem>
       end: BorderRadius.zero,
     ).animate(_controller);
 
-    _isExpanded = PageStorage.of(context)?.readState(context) as bool ??
```
shihaohong, line 73: @rami-a @JoseAlba Any idea why PageStorage is used for this particular widget? It currently does not support state restoration, but it seems like it's unnecessarily used here.

rami-a: I don't actually recall. I think it's probably safe to remove it

rami-a (approved):
lgtm

PR #379, opened 2020-11-20
Restorable Material Code Samples - Part 2/?

shihaohong:
## Description

More State Restoration Demos:
- Toggle Buttons
- Cards
- Chips
- Data Tables (selected row indexes cannot be restored -- filed https://github.com/flutter/flutter/issues/70916)
- Grid Lists
- Lists
- Menus

## Todos

Bottom Sheet needs work in the Scaffold (filed https://github.com/flutter/flutter/issues/70915)
Dialogs need to support restorablePush (filed https://github.com/flutter/flutter/issues/70918)


lib/demos/material/menu_demo.dart
```diff
@@ -277,37 +280,80 @@ class _SimpleMenuDemoState extends State<_SimpleMenuDemo> {
...
+  final String restorationId;
   final void Function(String value) showInSnackBar;
 
   @override
   _ChecklistMenuDemoState createState() => _ChecklistMenuDemoState();
 }
 
-class _ChecklistMenuDemoState extends State<_ChecklistMenuDemo> {
-  List<CheckedValue> _checkedValues;
+class _RestorableCheckedValue extends RestorableValue<Set<int>> {
```
shihaohong, line 296: Wondering what you think about this @goderbauer and if you have any suggestions for improving this.

lib/demos/material/data_table_demo.dart
```diff
@@ -36,13 +51,23 @@ class _DataTableDemoState extends State<DataTableDemo> {
   ) {
     _dessertsDataSource._sort<T>(getField, ascending);
     setState(() {
-      _sortColumnIndex = columnIndex;
-      _sortAscending = ascending;
+      // [RestorableBool]'s value cannot be null, so -1 is used as a placeholder
```
shihaohong, line 54: @goderbauer This is also an interesting problem, since `RestorableBool` does not take a null value, but some Flutter widgets use `null` to represent something meaningful.

lib/demos/material/button_demo.dart
```diff
@@ -150,12 +150,55 @@ class _OutlinedButtonDemo extends StatelessWidget {
...
+
+  final String restorationId;
+
   @override
   _ToggleButtonsDemoState createState() => _ToggleButtonsDemoState();
 }
 
-class _ToggleButtonsDemoState extends State<_ToggleButtonsDemo> {
-  final isSelected = <bool>[false, false, false];
+class _RestorableBoolList extends RestorableValue<List<bool>> {
```
shihaohong, line 163: @goderbauer

lib/demos/material/cards_demo.dart
```diff
@@ -373,13 +368,27 @@ class TravelDestinationContent extends StatelessWidget {
 }
 
 class CardsDemo extends StatefulWidget {
-  const CardsDemo();
+  const CardsDemo({
+    this.restorationId,
+  });
+
+  final String restorationId;
```
rami-a, line 375: Does the restorationId need to be a parameter to the constructor? Could it just be a const for each demo instead?

rami-a (approved):
LGTM but I think the restoration ID doesn't really need to be a constructor param on these demos and could be a const instead.

lib/demos/material/button_demo.dart
goderbauer: This only works if a new list is assigned to the property every time.

As an alternative, could _ToggleButtonsDemoState just have a List<RestorableBool> with three instances of RestorableBool in it? That may be simpler...

lib/demos/material/cards_demo.dart
goderbauer: If each demo is only ever used once, that would work just fine and could simplify things.

lib/demos/material/data_table_demo.dart
goderbauer: We can add an `RestorableBoolN` to the framework (and RestorableIntN, etc.) where the N stands for "this property is nullable". Ian and I discussed that as an option before. Feel free to open a PR on framework for that, if you need it.

lib/demos/material/menu_demo.dart
goderbauer: My suggestion would be to extend RestorableProperty directly and write a class that looks something like this:

```dart
class _RestorableCheckedValues extends RestorableProperty<List<int>> {
 Set<CheckedValue> _checked = Set<>();

  void check(CheckedValue value) {
    _checked.add(value);
   notifyListeners();
  }

  void uncheck(CheckedValue value) {
    _checked.remove(value);
   notifyListeners();
  }
  
  void isChecked(CheckedValue value) {
    return _checked.contains(value);
  }

  void initWithValue(Set<CheckedValue> a) {
    _checked = a;
  }

  @override
  Object toPrimitives() {
    return _checked.map((a) => e.index).toList();
  }

  @override
  Set<int> fromPrimitives(Object data) {
    final checkedValues = data as List<dynamic>;
    return Set.from(checkedValues.map<int>((dynamic id) => CheckedValues.values[id as int]));
  }

}
```

(this code probably doesn't compile, but you get the idea)

That way, you get a nice API to use within _ChecklistMenuDemoState.

Looking at the code below, instead of check/uncheck you could also have a simple toggle method. And you probably need to exposes a map function which it takes a callback and iterates over the values in _checked.

PR #389, opened 2020-12-16
[State Restoration] Shrine

shihaohong:
## Description

Adds State Restoration for Shrine mini app.

### Routes issue
/cc @goderbauer I'm not sure what's the problem here and I was debugging this for quite a while, but every time I try to restore the Shrine mini app, I get the following error message: 

Screen Shot 2020-12-16 at 11 35 34 AM
(had to use a screenshot because VSCode is giving me trouble with highlighting/copy-pasting the error)

Some of the lines may be a little off because I put some print statements in to try and debug and am quite lost. The setup of this app is similar to the Reply app, so I'm surprised that this one is acting up in this way. Any ideas?

To reproduce:
1. Open the Gallery app on an Android device
2. Swipe the carousel to the Shrine app card and tap it.
3. On the Shrine app home page, kill the activity and reopen the app. 

### Hot reload issue
The app seems to be state restorable now. However, I somehow broke hot reload with this PR... Any change in the code for `gallery/studies/shrine/app.dart` while on any screen in the gallery results in the following error message:
```
Unhandled exception:
root::package:gallery/studies/shrine/app.dart::_RestorableAppStateModel::@methods::package:flutter/src/widgets/restoration.dart::_debugAssertNotDisposed is already bound
#0      CanonicalName.bindTo (package:kernel/canonical_name.dart:192:7)
#1      Class.computeCanonicalNames (package:kernel/ast.dart:1201:51)
#2      Library.computeCanonicalNames (package:kernel/ast.dart:559:14)
#3      Component.computeCanonicalNamesForLibrary (package:kernel/ast.dart:10821:13)
#4      Component.computeCanonicalNames (package:kernel/ast.dart:10783:7)
#5      sortComponent (package:vm/kernel_front_end.dart:679:13)
#6      FrontendCompiler.writeDillFile (package:frontend_server/frontend_server.dart:656:5)
#7      FrontendCompiler.recompileDelta (package:frontend_server/frontend_server.dart:766:13)
<asynchronous suspension>

the Dart compiler exited unexpectedly.
Exited (1)
```

I'm not really sure what's going on. The app works fine otherwise.


lib/studies/shrine/app.dart
```diff
@@ -112,32 +112,36 @@ class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
...
-          theme: shrineTheme.copyWith(
-            platform: GalleryOptions.of(context).platform,
+    return RestorationScope(
+      restorationId: 'shrineAppState',
+      child: ScopedModel<AppStateModel>(
+        model: _model,
+        child: WillPopScope(
+          onWillPop: _onWillPop,
+          child: MaterialApp(
+            restorationScopeId: 'shrineApp',
```
shihaohong, line 122: The error is shows up with the addition of this line.

lib/studies/shrine/app.dart
```diff
@@ -112,32 +112,36 @@ class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
...
+            title: 'Shrine',
+            debugShowCheckedModeBanner: false,
+            initialRoute: ShrineApp.loginRoute,
+            onGenerateInitialRoutes: (_) {
+              return [
+                MaterialPageRoute<void>(
+                  builder: (context) => const LoginPage(),
```
goderbauer, line 129: The problem here is that the route(s) returned by `onGenerateInitialRoutes` do not have `Route.settings.name` set. This means, they cannot be restored and you end up with an empty set of routes to restore. We should provide a better error here. I'll file an issue for it.

In this case, the fix is really simple: Just remove the `onGenerateInitialRoutes` argument altogether. The app is already specifying an `initialRoute: ShrineApp.loginRoute`, which when looked up in the `routes` table below produces a route with a LoginPage - that's exactly the same thing that onGenerateInitialRoutes is doing. So instead of specifying it, we can just rely on the defaults.

goderbauer: Filed https://github.com/flutter/flutter/issues/72453.

shihaohong:
Everything seems to be working now. However, I somehow broke hot reload with this PR... Any change anywhere on any screen results in the following error message:
```
Connecting to VM Service at ws://127.0.0.1:63403/qOev_-I0w70=/ws
Unhandled exception:
root::package:gallery/studies/shrine/app.dart::_RestorableAppStateModel::@methods::package:flutter/src/widgets/restoration.dart::_debugAssertNotDisposed is already bound
#0      CanonicalName.bindTo (package:kernel/canonical_name.dart:192:7)
#1      Class.computeCanonicalNames (package:kernel/ast.dart:1201:51)
#2      Library.computeCanonicalNames (package:kernel/ast.dart:559:14)
#3      Component.computeCanonicalNamesForLibrary (package:kernel/ast.dart:10821:13)
#4      Component.computeCanonicalNames (package:kernel/ast.dart:10783:7)

```

shihaohong:
It turns out that it was because I was using `RestorationMixin` when a RestorationScope was not defined for the app. Wrapping the studies in a `RestorationScope` did the trick. I also removed extraneous uses of `RestorationScope`s in this PR.

shihaohong:
I made a small reproducible test app with the hot reload crash: https://github.com/shihaohong/state_restoration_provider_crash

lib/studies/shrine/app.dart
```diff
@@ -93,6 +129,15 @@ class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
 
   @override
   Widget build(BuildContext context) {
+    if (_isExpandingControllerCompleted.value !=
```
guidezpl, line 132: Using a statusListener for this feels a bit clunky. Could there be a cleaner way to do this?

guidezpl (approved):
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
                        
```

Merging should be held pending the hot reload crash fix

lib/studies/shrine/app.dart
shihaohong: I've been thinking about this too... Let me see if I can come up with a better API for this, here and everywhere else tab controller is being used.

shihaohong:
Filed an issue at https://github.com/dart-lang/sdk/issues/44523 and https://github.com/flutter/flutter/issues/72700. We'll see what the Dart folks say and see if they can identify the root cause.

lib/studies/shrine/app.dart
shihaohong: I refactored this out into its own class here: https://github.com/flutter/gallery/pull/389/files#diff-4576115e6281d08372fae69eebb7d35cff0bc534481112f35d3825370064b6abR32

Let me know what you think or if you have any improvement suggestions!

shihaohong:
According to https://github.com/flutter/flutter/issues/62916, this seems like a Dart issue unrelated to state restoration. Should we still hold off until that issue is fixed before moving forward with this PR?

lib/studies/shrine/app.dart
```diff
@@ -28,18 +29,76 @@ class ShrineApp extends StatefulWidget {
   _ShrineAppState createState() => _ShrineAppState();
 }
 
-class _ShrineAppState extends State<ShrineApp> with TickerProviderStateMixin {
+class _RestorableAnimationValue extends RestorableDouble {
+  _RestorableAnimationValue(double defaultValue) : super(defaultValue);
```
guidezpl, line 33: nit: `initialValue`

lib/studies/shrine/app.dart
guidezpl: Nice! For clarity, I wonder if `setAnimationController` would be better named as `registerStatusListener` or something similar

guidezpl:
If unrelated, this seems fine to merge

PR #390, opened 2020-12-16
State restorable fortnightly

shihaohong:
Just a few `ListView`s. The horizontal ListView is fine, however, I noticed that the vertical ListView's scroll offset isn't properly restored. Also, the asset images flicker upon state restoration. A similar issue is happening for the Reply mini app as well.

My guess is that the images need to be loaded when the app is restored, but at the first layout, they're not available and result in the offset being restored incorrectly. ~If this is really the case, is there a way to wait for the images to load before building the ListView? I stumbled upon [this flutter issue](https://github.com/flutter/flutter/issues/26127) while looking for clues.~

Another solution could be to put a placeholder in place before the asset image is available, but it's weird because [FadeInImagePlaceholder](https://github.com/flutter/gallery/blob/master/lib/layout/image_placeholder.dart#L14) should already be doing this? I'm not too familiar with images so I'm not exactly sure what's going on here.

Edit: I ended up modifying the code so that it uses a placeholder when the asset image is not loaded yet. PTAL

### DIFF
--- a/lib/layout/image_placeholder.dart
+++ b/lib/layout/image_placeholder.dart
@@ -11,7 +11,7 @@ import 'package:flutter/material.dart';
 /// to specify a widget as a [placeholder], instead of just an [ImageProvider].
 /// It also lets you override the [child] argument, in case you want to wrap
 /// the image with another widget, for example an [Ink.image].
-class FadeInImagePlaceholder extends StatelessWidget {
+class FadeInImagePlaceholder extends StatefulWidget {
   const FadeInImagePlaceholder({
     Key key,
     @required this.image,
@@ -55,22 +55,47 @@ class FadeInImagePlaceholder extends StatelessWidget {
   final BoxFit fit;
 
   @override
+  _FadeInImagePlaceholderState createState() => _FadeInImagePlaceholderState();
+}
+
+class _FadeInImagePlaceholderState extends State<FadeInImagePlaceholder> {
+  bool _isLoaded = false;
+
+  @override
+  void initState() {
+    // Once the image is loaded, we no longer need to show a placeholder
+    widget.image
+        .resolve(ImageConfiguration.empty)
+        .addListener(ImageStreamListener(
+      (imageInfo, synchronousCall) {
+        setState(() {
+          _isLoaded = true;
+        });
+      },
+    ));
+    super.initState();
+  }
+
+  @override
   Widget build(BuildContext context) {
     return Image(
-      image: image,
-      excludeFromSemantics: excludeFromSemantics,
-      width: width,
-      height: height,
-      fit: fit,
+      image: widget.image,
+      excludeFromSemantics: widget.excludeFromSemantics,
+      width: widget.width,
+      height: widget.height,
+      fit: widget.fit,
       frameBuilder: (context, child, frame, wasSynchronouslyLoaded) {
         // We only need to animate the loading of images on the web. We can
         // also return early if the images are already in the cache.
         if (wasSynchronouslyLoaded || !kIsWeb) {
-          return this.child ?? child;
+          if (!_isLoaded) {
+            return widget.placeholder;
+          }
+          return widget.child ?? child;
         } else {
           return AnimatedSwitcher(
-            duration: duration,
-            child: frame != null ? this.child ?? child : placeholder,
+            duration: widget.duration,
+            child: frame != null ? widget.child ?? child : widget.placeholder,
           );
         }
       },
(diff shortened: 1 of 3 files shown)


lib/layout/image_placeholder.dart
```diff
@@ -54,23 +54,48 @@ class FadeInImagePlaceholder extends StatelessWidget {
...
+    widget.image
+        .resolve(ImageConfiguration.empty)
+        .addListener(ImageStreamListener(
+      (imageInfo, synchronousCall) {
+        setState(() {
+          _isLoaded = true;
+        });
+      },
+    ));
+    super.initState();
```
rami-a, line 76: I believe super should be called first

lib/studies/fortnightly/app.dart
```diff
@@ -24,6 +24,7 @@ class FortnightlyApp extends StatelessWidget {
         ? const _FortnightlyHomeDesktop()
         : const _FortnightlyHomeMobile();
     return MaterialApp(
+      restorationScopeId: 'appNavigator',
```
rami-a, line 27: Should the ID be something like `fortnightlyApp`?

rami-a (approved):

lib/studies/fortnightly/app.dart
shihaohong: It shouldn't matter since only one mini-app can be open at any given time so there won't be a conflict. I'll change it anyway since it's pretty quick and may be clearer.

lib/layout/image_placeholder.dart
```diff
@@ -54,23 +54,48 @@ class FadeInImagePlaceholder extends StatelessWidget {
...
+
+class _FadeInImagePlaceholderState extends State<FadeInImagePlaceholder> {
+  bool _isLoaded = false;
+
+  @override
+  void initState() {
+    super.initState();
+    // Once the image is loaded, we no longer need to show a placeholder
+    widget.image
+        .resolve(ImageConfiguration.empty)
```
goderbauer, line 69: This will wait for the asset with the given configuration, which is not necessarily the one that ends up being shown on screen?

lib/layout/image_placeholder.dart
```diff
@@ -54,23 +54,48 @@ class FadeInImagePlaceholder extends StatelessWidget {
...
+      excludeFromSemantics: widget.excludeFromSemantics,
+      width: widget.width,
+      height: widget.height,
+      fit: widget.fit,
       frameBuilder: (context, child, frame, wasSynchronouslyLoaded) {
         // We only need to animate the loading of images on the web. We can
         // also return early if the images are already in the cache.
         if (wasSynchronouslyLoaded || !kIsWeb) {
-          return this.child ?? child;
+          if (!_isLoaded) {
```
goderbauer, line 91: Why do we need the extra _isLoaded flag? How does it differ from wasSynchronouslyLoaded?

lib/layout/image_placeholder.dart
```diff
@@ -54,23 +54,48 @@ class FadeInImagePlaceholder extends StatelessWidget {
...
+            return widget.placeholder;
```
goderbauer, line 92: Wouldn't we also want to switch from the placeholder to the actual image in an animated way here as well?

goderbauer:
> I noticed that the vertical ListView's scroll offset isn't properly restored

Is that because of the assets? Or is this a separate problem?

lib/layout/image_placeholder.dart
```diff
@@ -54,23 +54,48 @@ class FadeInImagePlaceholder extends StatelessWidget {
...
+          }
+          return widget.child ?? child;
```
goderbauer, line 94: I think we can simplify this a lot and avoid the bug mentioned above where we resolve it for an incorrect configuration by removing the _isLoaded flag altogether and just changing this code to:

```suggestion
          if (frame == null) {
            return widget.placeholder;
          }
          return widget.child ?? child;
```

FadeInImagePlaceholder can be a StatelessWidget again after that.
Not sure if we want the transition from placeholder to real image to be animated here as well or not (I am leaning towards animation).

goderbauer: If we want it animated, we could simply fo this:

```dart
        if (wasSynchronouslyLoaded) {
          return this.child ?? child;
        } else {
          return AnimatedSwitcher(
            duration: duration,
            child: frame != null ? this.child ?? child : placeholder,
          );
        }
```

(And refactor the widget to be stateless again, of course).

shihaohong: I originally tried what you suggested in the most recent comment, and when I did, it didn't work -- it works fine now. 

I think what was happening is I was testing state restoration while testing this part of the app and was getting confused -- Every time the app is killed and restored, the app ends up in the state at which `flutter run` was called, not including any of the changes I made to the gallery code afterwards.

PR #391, opened 2020-12-16
State Restorable starter app

shihaohong:
Not much to do here -- Once drawers are state restorable in the framework (see https://github.com/flutter/flutter/issues/70925), this should automatically pick that up. Otherwise, this will prepare the starter app for any future state restoration additions to the mini app.

Related to https://github.com/flutter/flutter/issues/62916

### DIFF
--- a/lib/studies/starter/app.dart
+++ b/lib/studies/starter/app.dart
@@ -17,6 +17,7 @@ class StarterApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      restorationScopeId: 'appNavigator',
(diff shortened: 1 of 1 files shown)


lib/studies/starter/app.dart
```diff
@@ -17,6 +17,7 @@ class StarterApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      restorationScopeId: 'appNavigator',
```
rami-a, line 20: Should the ID be something like `starterApp`?

rami-a (approved):

PR #392, opened 2020-12-16
State Restorable Crane app

shihaohong:
State restorable Crane app, except for a single use of `StaggeredGridView`, which I've opened a PR for (https://github.com/letsar/flutter_staggered_grid_view/pull/148/).

https://github.com/flutter/flutter/issues/62916

### DIFF
--- a/lib/studies/crane/app.dart
+++ b/lib/studies/crane/app.dart
@@ -19,7 +19,7 @@ class CraneApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
-      restorationScopeId: 'appNavigator',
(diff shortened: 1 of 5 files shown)


lib/studies/crane/app.dart
```diff
@@ -19,6 +19,7 @@ class CraneApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return MaterialApp(
+      restorationScopeId: 'appNavigator',
```
rami-a, line 22: I know you mentioned this ID doesn't matter but for consistency this should be craneApp I guess

rami-a (approved):
LGTM

PR #403, opened 2021-01-04
[State Restoration] TabController and AnimationController usage

shihaohong:
I was originally attempting to set restoration state in the `build` method in a hacky way. In reality, using `restoreState` seems like the better approach.


guidezpl (approved):
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
                        
```
Nice!

PR #404, opened 2021-01-06
[State Restoration] DataTable selected rows state

shihaohong:
Turns out state restoration has to be done on the user side of things. In `PaginatedDataTable`s, `DataRow`s are created by the `DataTableSource` class. This PR adds state restoration to the selection state of the `PaginatedDataTable`.

Closes https://github.com/flutter/flutter/issues/70916


lib/demos/material/data_table_demo.dart
```diff
@@ -17,11 +17,39 @@ class DataTableDemo extends StatefulWidget {
   _DataTableDemoState createState() => _DataTableDemoState();
 }
 
+class _RestorableSelectedDessertIndices extends RestorableValue<Set<int>> {
```
goderbauer, line 20: This design makes it tempting to just modify the `values` set directly, which doesn't work because those updates will not be restored.

Instead of exposing the value set directly, I would write a custom RestorableProperty, which does not expose the set and instead has an interface like this:

```
class Foo extends ... {
  setSelection(Set<int> selectedIndexes);
  isSelected(int index);
}
```

shihaohong: Done

shihaohong:
Ready for another round of review. I also fixed an issue where the selected count of the data table wasn't being restored correctly.

goderbauer (approved):
LGTM

PR #409, opened 2021-01-27
[State Restoration] Material Dialogs, Cupertino Dialogs, CupertinoTabScaffold/TabView, CupertinoTextFields

shihaohong:
Related to https://github.com/flutter/flutter/issues/62916.

Made the following demos state restorable: 
1) Material Dialogs
2) Cupertino Dialogs (~except ActionSheets/Modals, those will come when the framework changes (https://github.com/flutter/flutter/pull/74805) have landed~)
3) CupertinoTabScaffold/TabView
4) CupertinoTextFields


lib/demos/cupertino/cupertino_alert_demo.dart
```diff
@@ -29,8 +29,63 @@ class CupertinoAlertDemo extends StatefulWidget {
...
+    registerForRestoration(_alertDialogRoute, 'alert_demo_dialog_route');
+    registerForRestoration(
+        _alertWithTitleDialogRoute, 'alert_with_title_press_demo_dialog_route');
+    registerForRestoration(_alertWithButtonsDialogRoute,
+        'alert_with_title_press_demo_dialog_route');
+    registerForRestoration(_alertWithButtonsOnlyDialogRoute,
+        'alert_with_title_press_demo_dialog_route');
+  }
+
+  void setSelectedValue(String value) {
```
goderbauer, line 55: make private?

```suggestion
  void _setSelectedValue(String value) {
```

lib/demos/cupertino/cupertino_alert_demo.dart
```diff
@@ -29,8 +29,63 @@ class CupertinoAlertDemo extends StatefulWidget {
...
+    setState(() {
```
goderbauer, line 56: In the old code, this only happened when `value != null` - should we keep that behavior?

lib/demos/cupertino/cupertino_alert_demo.dart
```diff
@@ -50,118 +105,120 @@ class _CupertinoAlertDemoState extends State<CupertinoAlertDemo> {
     return '';
   }
 
-  void _showDemoDialog({BuildContext context, Widget child}) {
-    showCupertinoDialog<String>(
+  static Route<String> _alertDemoDialog(
+      BuildContext context, Object arguments) {
```
goderbauer, line 109: nit (although I am not super familiar with the code formatting style guide here): I'd add a trailing comma behind `arguments` and format this in multiple lines.

lib/demos/cupertino/cupertino_alert_demo.dart
```diff
@@ -50,118 +105,120 @@ class _CupertinoAlertDemoState extends State<CupertinoAlertDemo> {
...
+          actions: [
+            CupertinoDialogAction(
+              child: Text(
+                GalleryLocalizations.of(context).cupertinoAlertDiscard,
+              ),
+              isDestructiveAction: true,
+              onPressed: () {
+                Navigator.of(
+                  context,
+                  rootNavigator: true,
```
goderbauer, line 124: Why `rootNavigator: true` here and elsewhere? Isn't the dialog just pushed into the closest enclosing dialog?

lib/demos/material/dialog_demo.dart
```diff
@@ -35,101 +90,148 @@ class DialogDemo extends StatelessWidget {
...
         .copyWith(color: theme.textTheme.caption.color);
-    _showDemoDialog<String>(
+
+    final themes = InheritedTheme.capture(
+      from: context,
+      to: Navigator.of(
+        context,
+        rootNavigator: true,
+      ).context,
+    );
```
goderbauer, line 107: Why do we need to capture these themes? My understanding is that is is capturing the themes between the Navigator that will own the route and the root Navigator. However, everything above the Navigator that will own the route should already be visible to the route, so there shouldn't be a need to capture?

lib/demos/material/dialog_demo.dart
```diff
@@ -35,101 +90,148 @@ class DialogDemo extends StatelessWidget {
...
+    return DialogRoute<String>(
       context: context,
-      child: AlertDialog(
-        content: Text(
-          GalleryLocalizations.of(context).dialogDiscardTitle,
-          style: dialogTextStyle,
+      themes: themes,
+      builder: (context) => ApplyTextOptions(
+        child: Theme(
+          data: Theme.of(context),
```
goderbauer, line 114: Why this? Isn't this just wrapping everything in the same ThemeData that children could access via `Theme.of` even if it wasn't re-wrapped like this?

lib/demos/material/dialog_demo.dart
goderbauer: Same question for the instances below.

lib/demos/material/dialog_demo.dart
goderbauer: Same question for the instances below.

lib/demos/material/dialog_demo.dart
shihaohong: I was just copying the default behavior without attempting to modify anything, but you're right, it's unnecessary. Removed it!

lib/demos/cupertino/cupertino_alert_demo.dart
shihaohong: Yeah, you're right. Again, I just copied over the behavior that was in the existing code, but it's not necessary so I removed it

lib/demos/cupertino/cupertino_alert_demo.dart
shihaohong: Yeah, sorry about that. I think I confused myself moving between NNBD and non-NNBD codebases, since the gallery is still pre-NNBD

shihaohong:
Addressed the code review and added action sheet state restoration, since that has made it into the `flutter/flutter` master branch. PTAL

goderbauer (approved):
LGTM

shihaohong:
Odd, the `web-benchmark` tests are failing on the windows check for reasons unrelated to the PR. I'll look into it more later, but I'm not really sure what's going on at first glance.

shihaohong:
The web benchmark tests have been fixed with https://github.com/flutter/packages/pull/281 and is unrelated to this change.

shihaohong:
The golden tests are failing, regardless of whether they have been updated or not. In any case, these changes should not have resulted in the golden failures seen since the state restoration changes do not contain any UI changes.

shihaohong:
Merging with the failed run goldens, since that doesn't seem related to the changes made in this PR

Issue #411, opened 2021-01-28
Web Benchmarks failing


shihaohong:
Unsure what started this issue, but the web benchmarks seemed to have stopped working.

I submitted an almost empty PR to reproduce: https://github.com/flutter/gallery/pull/410. Thinking it might've been some dependency problem issue, I ran `flutter pub get` which updated the version for the transitive `webdriver` dependency. This caused the Mac bot to fail similarly to how the Windows bot was failing when I first spotted the issue (see my next comment for what shows on the Windows bot without any changes to `pubspec.lock`).

cc/ @pennzht

This is the resulting error message: 
```
00:10 +0: D:\a\gallery\gallery\code\test\benchmarks_test.dart: Can run a web benchmark                                                                                                                 
Starting web benchmark tests ...
Launching Chrome.
Launching Chrome...
Waiting for the benchmark to report benchmark profile.
[CHROME]: 
[CHROME]: DevTools listening on ws://127.0.0.1:10000/devtools/browser/81cac324-475c-443b-a818-129f5be37024
Connecting to DevTools: ws://localhost:10000/devtools/page/F83A9AA35D3498E85674D341B9613979
Connected to Chrome tab:  (http://localhost:9999/index.html)
Launching benchmark "gallery_v2_studies_perf"
[APP] Warming up.
[APP] Warm-up finished.
[APP] ==== List of demos to be run ====
[APP] shrine@study
[APP] rally@study
[APP] crane@study
[APP] fortnightly@study
[APP] reply@study
[APP] starter@study
[APP] ==== End of list of demos to be run ====
[APP] Running demo "shrine@study"
[APP] Running demo "rally@study"
[APP] Running demo "crane@study"
[APP] Running demo "fortnightly@study"
[APP] Running demo "reply@study"
[APP] Running demo "starter@study"
[APP] All demos finished.
Shell: Failed to interpret the Chrome trace contents. The trace was saved in ./chrome-trace.json
Extracted 269 measured frames.
Skipped 32 non-measured frames.

02:50 +0 -1: D:\a\gallery\gallery\code\test\benchmarks_test.dart: Can run a web benchmark [E]                                                                                                          
  NoSuchMethodError: The getter 'tdur' was called on null.
  Receiver: null
  Tried calling: tdur
  dart:core                                       Object.noSuchMethod
  package:web_benchmarks/src/browser.dart 452:15  _computeAverageDuration.<fn>
  dart:_internal                                  ListIterable.fold
[...]

PR #412, opened 2021-01-28
Disable web benchmarks

shihaohong:
Seems like web benchmarks started failing unusually. It's not clear what's changed, but I want to skip the test for now to unblock other work, if that's okay.

GitHub issue: https://github.com/flutter/gallery/issues/411


rami-a (approved):
LGTM, this is blocking several PRs now so I think we should skip it until we can fix it

clocksmith (approved):

rami-a:
Actually https://github.com/flutter/packages/pull/281 might fix this instead

PR #425, opened 2021-02-08
[Reply] Cap body at 1340 for desktop and tablet

### LINKED ISSUE #298: [Reply] Cap the desktop body width to 1340
Currently, the width extends as wide as it can be with constant margins, but that body with should be capped at 1340 for better text readability.

Renzo-Olivares:
This change caps the body of the reply material study to 1340 for better text readability.

desktop_cap

Closes #298


lib/studies/reply/adaptive_nav.dart
```diff
@@ -241,12 +241,23 @@ class _DesktopNavState extends State<_DesktopNav>
...
             },
           ),
-          const VerticalDivider(thickness: 1, width: 1),
+          const VerticalDivider(thickness: 1, width: 2),
           Expanded(
-            child: _SharedAxisTransitionSwitcher(
-              defaultChild: _MailNavigator(
-                child: MailboxBody(key: widget.inboxKey),
-              ),
+            child: Row(
```
rami-a, line 246: I think you should be able to achieve this without using a Row. Perhaps replace the Row and Flexible widgets with a Center widget?

lib/studies/reply/adaptive_nav.dart
```diff
@@ -241,12 +241,23 @@ class _DesktopNavState extends State<_DesktopNav>
               );
             },
           ),
-          const VerticalDivider(thickness: 1, width: 1),
+          const VerticalDivider(thickness: 1, width: 2),
```
rami-a, line 244: Whats the reason for this change?

rami-a (commented):
Thanks @Renzo-Olivares 

Could you please take a look at the comments, and also update this branch against the latest master as I have fixed broken tests

lib/studies/reply/adaptive_nav.dart
Renzo-Olivares: Thanks for the suggestion that works!

lib/studies/reply/adaptive_nav.dart
Renzo-Olivares: white_lines

I increased the width to remedy some artifacts (white lines shown above). From my observation, they seem to be a result of the navigation rail body not expanding to its max-width like it previously did before this change.

rami-a (approved):
LGTM

clocksmith:
LGTM, assuming there are still constant margins for when the width is less than 1340

lib/studies/reply/adaptive_nav.dart
Renzo-Olivares: These lines still show up if I don't have my browser window maximized. I'll look into it some more.

lib/studies/reply/adaptive_nav.dart
```diff
@@ -241,11 +241,17 @@ class _DesktopNavState extends State<_DesktopNav>
...
           Expanded(
-            child: _SharedAxisTransitionSwitcher(
-              defaultChild: _MailNavigator(
-                child: MailboxBody(key: widget.inboxKey),
+            child: Center(
+              child: ConstrainedBox(
+                constraints:
+                    BoxConstraints(maxWidth: widget.extended ? 1340 : 800),
```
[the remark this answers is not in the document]
Renzo-Olivares: Yeah, you're right. I misunderstood the purpose of the cap and made it consistent for both desktop and tablet. I changed it back to match the original issue, also the margins were not constant when the width was under 1340 with the added 800 cap.

PR #426, opened 2021-02-09
Update all packages

guidezpl:
First step to migrate the gallery to null safety is to update dependencies.

`flutter pub upgrade`
`flutter pub upgrade --major-versions`
`flutter pub upgrade --null-safety`

There is only one package ( `web_benchmarks`) that isn't null safe yet, but it doesn't block the migration.


pennzht:
@guidezpl 

You can depend on `web_benchmarks 0.0.4` now

pennzht (approved):
**Thank you!**
**LGTM**

PR #436, opened 2021-02-19
Roll path_provider to a version compatible with the latest Dart SDK

jason-simmons:


pubspec.yaml
```diff
@@ -37,6 +37,7 @@ dev_dependencies:
   args: ^2.0.0
   grinder: ^0.9.0-nullsafety.0
   path: ^1.8.0
+  path_provider: ^2.0.0-nullsafety.1
```
zanderso, line 40: I was expecting this to be a dependency override. Why is it a dev_dependency?

zanderso (approved):
lgtm w/ explanation of why a dev_dependency rather than a dependency override.

pubspec.yaml
jason-simmons: Updated this to an override

PR #439, opened 2021-02-22
[DO NOT MERGE] Update macos version for golden test

shihaohong:


shihaohong:
I wanted to see if this was a result of using Catalina vs Big Sur, but it turns out that GitHub only has Big Sur bots available for preview: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners

PR #445, opened 2021-02-28
[Reply] Fix lines around MailBody when width cap is exceeded

Renzo-Olivares:
The `SharedAxisTransition` widget has a property called `fillColor` which is the background color used during the transition. This fillColor looks like it also outlines the widget it contains, so here I change it to match the background of the background so they are no longer visible.

This change fixes the bug introduced in #425. Where faint lines would surround the MailBody (see below) when exceeding the new 1340 width cap.

Before|After
---|---
capexample|fixeddddddd


guidezpl (approved):
LGTM without the goldens

guidezpl:
Thanks!

PR #455, opened 2021-03-10
Remove GTK backend check on Linux (Wayland is now supported)

wmww:
Same as #454, except might work better because I've signed the CLA


guidezpl:
Thanks for contributing! I recreated the linux subdirectory and included changes from https://github.com/flutter/gallery/pull/458

guidezpl (approved):

PR #458, opened 2021-03-12
Make the GApplication non-unique

Saviq:
This avoids the DBus requirement.


guidezpl:
Thanks for contributing! Your change is part of https://github.com/flutter/gallery/pull/455, so I'll close this

PR #460, opened 2021-03-22
[WIP] Found localization mistakes

shihaohong:
These are all mistakes I found from working on an improvement to the `gen-l10n` tool. I'm not sure what the best way to resolve all of them is, but creating this PR was the way I tracked and discovered these issues. They'll need to be fixed in the google translation console.

cc/ @rami-a


rami-a:
Ah, I've seen stuff like this happen before if the translators get confused with string parameters. You can request corrections in the translation tool

shihaohong:
I just submitted all the change requests to the translation console. Hopefully, these changes will be made by the next time the localizations are requested the normal way!

PR #461, opened 2021-03-22
Build `crane` demo as deferred component

GaryQian:
This opts into deferred components (go/deferred-components-instructions) and moves the `crane` demo to be deferred loaded.

This also updates the version of gradle, dart, and kotlin to be in sync with flutter master, and adds play core as a dependency.

Almost every section of the app can actually be split off as a deferred component, but this PR only does this for `crane`, which can act as an example for how to use this feature on a larger app.


android/build.gradle
```diff
@@ -1,12 +1,12 @@
 buildscript {
-    ext.kotlin_version = '1.2.71'
+    ext.kotlin_version = '1.3.50'
```
GaryQian, line 2: this particular upgrade is not necessary for this PR, but is done anyways to keep up to date with master flutter.

lib/deferred_widget.dart
```diff
@@ -68,6 +70,42 @@ class _DeferredWidgetState extends State<DeferredWidget> {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Container(
+        child: Column(
+          children: <Widget>[
+            Text('$name is installing.', style: Theme.of(context).textTheme.headline4),
+            Container(height: 10),
+            Text('$name is a deferred component which are downloaded and installed at runtime.', style: Theme.of(context).textTheme.bodyText1),
```
rami-a, line 92: You'll need to run dartfmt on these changes I believe

lib/deferred_widget.dart
```diff
@@ -68,6 +70,42 @@ class _DeferredWidgetState extends State<DeferredWidget> {
...
+        ),
+        decoration: BoxDecoration(
+          color: Colors.grey[700],
+          border: Border.all(
+            width: 20.0,
+            color: Colors.grey[700],
+          ),
+          borderRadius: const BorderRadius.all(Radius.circular(10))
+        ),
+        width: 250.0,
```
rami-a, line 107: Nit: Drop the `.0` from the double. That's what we prefer for the gallery

pubspec.yaml
```diff
@@ -46,6 +46,54 @@ dependency_overrides:
   path_provider: ^2.0.0-nullsafety.1
 
 flutter:
+  deferred-components:
```
rami-a, line 49: By making crane deferred, what happens on platforms that do not support deferred components?

rami-a (approved):
LGTM but have a few questions/comments

pubspec.yaml
GaryQian: Nothing, they behave/build as normal. The tooling will handle assets declared here as regular assets defined in the `assets` section. Deferred builds only happen when building release appbundles.

jimmycanosa: This is the error I am always getting. how to fix the deferred-components? this is my error:

[gallery] flutter create --ios-language swift --android-language kotlin .
Error detected in pubspec.yaml:
Unexpected child "deferred-components" found under "flutter".
Please correct the pubspec.yaml file at .\pubspec.yaml
exit code 1

--

[gallery] flutter create --ios-language swift --android-language kotlin .
Error detected in pubspec.yaml:
Unexpected child "deferred-components" found under "flutter".
Please correct the pubspec.yaml file at .\pubspec.yaml
exit code 1

--

[gallery] flutter pub get
Error detected in pubspec.yaml:
Unexpected child "deferred-components" found under "flutter".
Please correct the pubspec.yaml file at D:\Flutter\gallery\pubspec.yaml
exit code 1

GaryQian: It looks like your tool is outdated? Try `flutter upgrade` after switching to beta or master channel. The new tooling has support for deferred components and the pubspec validator should recognize the new setting.

wer-mathurin:
@GaryQian Seems this will resolve the problem delivering an app with lot of assets.  I need to dive in the code to see how it work, but any chance that  the official flutter build documentation will be updated with this nice addition.  I think the common use case is serving deferred assets.  In our case we have more than 1500 images, so we end up having just above the 150MB limit.  We can reduce the file size, but we want to keep the images with different quality to serve mobile and web app.

GaryQian:
The docs at https://flutter.dev/docs/perf/deferred-components should include instructions on how to include your assets as deferred components!

The main difference between deferred code and deferred-assets-only is how you trigger the loading of the deferred component. Specifically, you should be calling the installDeferredComponent method on https://api.flutter.dev/flutter/services/DeferredComponent-class.html instead of the loadLibrary method.

PR #470, opened 2021-03-31
Use writeTimelineToFile

dnfield:


guidezpl (approved):
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
                        
```

Regarding the test, the updated goldens LGTM. You may want to temporarily disable the integration test for web due to https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/test.dart#L411

dnfield:
Added the goldens.

dnfield:
The integration test looks like it was broken by https://github.com/flutter/flutter/pull/74236 - @jiahaog

dnfield:
Oh. I don't have permission to land it while the status is broken haha.

dnfield:
Ok - the test in question isn't actually depending on dart:ui or package:integration_test, so I'm switching it to just use `dart test`, which works. Flutter is special casing testsin the `integeration_test` directory now.

jiahaog:
> The integration test looks like it was broken by flutter/flutter#74236 - @jiahaog

@dnfield correct me if I'm wrong, you mean https://github.com/flutter/flutter/pull/76200 for [this failure](https://github.com/flutter/gallery/runs/2239268420) right

dnfield:
Ahh yes, I think I pasted or blamed the wrong PR.

PR #473, opened 2021-04-12
Fix onKey handler return values

gspencergoog:
This fixes the return values for the `onKey` handlers used in the gallery code to use the newer `KeyEventResult` values, rather than bools. The use of the bools will soon be deprecated.

See also: https://github.com/flutter/flutter/pull/80294


rami-a (approved):
LGTM

gspencergoog:
Thanks for the review, @rami-a

In order for the formatting tests to pass, I had to include some files where the formatting was fixed: it seems some snuck past the check in previous PRs.

PR #476, opened 2021-04-16
[State Restoration] DatePicker, DateRangePicker, TextFormField

shihaohong:
As titled.

Also fixes an issue in the text field demo where entered information would disappear when scrolled away from because ListView  lazily loads its children, so any widget not in view would get tossed and recreated.


shihaohong:
Might need to wait for https://github.com/flutter/gallery/pull/477 to land before CI will pass. Seems like some formatting rules have changed and is causing the fresh `master` branch to fail

rami-a (approved):
LGTM

PR #480, opened 2021-04-20
Remove the AnimatedSize vsync parameter

jason-simmons:
This is needed in order to land https://github.com/flutter/flutter/pull/80554,
which will deprecate AnimatedSize.vsync

### DIFF
--- a/lib/pages/demo.dart
+++ b/lib/pages/demo.dart
@@ -405,7 +405,6 @@ class _GalleryDemoPageState extends State<GalleryDemoPage>
       );
     } else {
       section = AnimatedSize(
-        vsync: this,
         duration: const Duration(milliseconds: 200),
         alignment: Alignment.topCenter,
         curve: Curves.easeIn,
--- a/lib/studies/shrine/expanding_bottom_sheet.dart
+++ b/lib/studies/shrine/expanding_bottom_sheet.dart
@@ -132,8 +132,7 @@ double _getPeakPoint({double begin, double end}) {
   return begin + (end - begin) * _peakVelocityProgress;
 }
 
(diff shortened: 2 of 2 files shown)


lib/studies/shrine/expanding_bottom_sheet.dart
```diff
@@ -556,7 +556,6 @@ class _ExpandingBottomSheetState extends State<ExpandingBottomSheet>
       key: _expandingBottomSheetKey,
       duration: const Duration(milliseconds: 225),
       curve: Curves.easeInOut,
-      vsync: this,
```
rami-a, line 559: Would this class still need to be a `TickerProviderStateMixin`

lib/pages/demo.dart
```diff
@@ -405,7 +405,6 @@ class _GalleryDemoPageState extends State<GalleryDemoPage>
       );
     } else {
       section = AnimatedSize(
-        vsync: this,
```
rami-a, line 408: Would this class still need to be a `TickerProviderStateMixin`?

rami-a (approved):
LGTM but have a couple questions

lib/studies/shrine/expanding_bottom_sheet.dart
jason-simmons: Removed the `TickerProviderStateMixin`

lib/pages/demo.dart
jason-simmons: This is still needed because `_GalleryDemoPageState` is also creating an `AnimationController` that uses a `TickerProvider`

goderbauer (approved):
LGTM

Issue #471, opened 2021-04-06
Schedule for releaseing new versions


Mister-Hope:
The lastest release was on 18 Sep 2020.

I am wondering if a version taget with null safety could be realesed, And if a release schedule could be made (E.g.: Automatically release a version building with newest code on 1st of each month)

guidezpl:
That's a good suggestion! In my spare time, I've been working on automating releases. I have fastlane deployments left to do, after which the release cadence will become regular. As for null safety, the gallery actually hasn't yet been migrated.

guidezpl:
Closing since this is now set up, there will be more regular releases 🚀

PR #487, opened 2021-04-26
Create web deployment workflow

guidezpl:
Add workflow to deploy to Firebase hosting, with the ability to [run workflow manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) and specify which environment (`prod`, `staging` by default) to deploy to. 

Screenshot 2021-04-26 at 17 44 23

Remove workaround and enable deferred loading for localizations by default, since the hot reload l10n.yaml [issue](https://github.com/flutter/gallery/issues/357) no longer occurs.

Pin flutter-action action to v1.5.0 now that it is available.

Closes https://github.com/flutter/flutter/issues/69023
#471


rami-a (approved):
lgtm

PR #497, opened 2021-05-10
Removed obsolete references to ThemeData.buttonColor

HansMuller:
The ThemeData.buttonColor is no longer used. This change will enable deprecating it in https://github.com/flutter/flutter/pull/82196


darrenaustin (approved):
LGTM

Issue #504, opened 2021-05-19
cupertino_segmented_control_demo Container not work


zhhding:
file :cupertino_segmented_control_demo
row:87
child: children[currentSegment] = > child: children[currentSegment.value]

PR #506, opened 2021-05-19
Fix segmented control demo 

guidezpl:
Fixes https://github.com/flutter/gallery/issues/504

Also added a lower Flutter sdk lower bound


rami-a (approved):
Lgtm

PR #508, opened 2021-05-25
Apply upcoming flutter lints

guidezpl:
Apply most upcoming lints from https://flutter.dev/docs/release/breaking-changes/flutter-lints-package, but still rely on `pedantic` (can only include 1 in `analysis_options.yaml`)

`const`s over `final`, `const` constructors with `Key`, replacing unnecessary `Container`s with `SizedBox`, string interpolation, debug printing, subclass parameter naming


test/benchmarks_test.dart
```diff
@@ -22,7 +23,7 @@ final valueList = <String>[
 /// Tests that the Gallery web benchmarks are run and reported correctly.
 Future<void> main() async {
   test('Can run a web benchmark', () async {
-    print('Starting web benchmark tests ...');
+    stdout.writeln('Starting web benchmark tests ...');
```
rami-a, line 26: These changes seem odd, any idea why they happened?

rami-a (approved):

test/benchmarks_test.dart
guidezpl: Yeah, this it the lint https://dart-lang.github.io/linter/lints/avoid_print.html

Issue #523, opened 2021-06-16
Auto-Scrollbars on Desktop are disabled


Piinks:
By default on desktop, scrollbars are applied by the ScrollBehavior. This has been disabled in the gallery app. All vertical scrollables in the gallery need to be audited before enabling this feature.
Any that are vertical without a specified scroll controller automatically attach to the primary scroll controller. This causes issues when interacting with scrollbars because they are all trying to use the same scroll controller. To audit, the main vertical scrollable on a given page should use the primary scroll controller, any others should set `primary: false`. They will each get their own unique scroll controller for use with the scrollbar then.

--------

### Previous Context:

Running the gallery app on desktop produces an error related to the scrollbars that are being automatically applied by ScrollBehavior.buildScrollbar.

Some discussion: https://discord.com/channels/608014603317936148/608021234516754444/854798350246215751

Change that appears to have introduced the error: https://github.com/flutter/flutter/pull/81278

Stack trace: 

```
An Observatory debugger and profiler on macOS is available at: http://127.0.0.1:55818/-9YpxMWB5N8=/

══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following assertion was thrown during a scheduler callback:
The provided ScrollController is currently attached to more than one ScrollPosition.
The Scrollbar requires a single ScrollPosition in order to be painted.
When the scrollbar is interactive, the associated Scrollable widgets must have unique
ScrollControllers. The provided ScrollController must be unique to a Scrollable widget.

When the exception was thrown, this was the stack:
#0      RawScrollbarState._checkForValidScrollPosition.<anonymous closure> (package:flutter/src/widgets/scrollbar.dart:1101:9)
#1      RawScrollbarState._checkForValidScrollPosition (package:flutter/src/widgets/scrollbar.dart:1125:6)
#2      RawScrollbarState._maybeTriggerScrollbar.<anonymous closure> (package:flutter/src/widgets/scrollbar.dart:1042:9)
#3      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1143:15)
#4      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1088:9)
#5      SchedulerBinding.scheduleWarmUpFrame.<anonymous closure> (package:flutter/src/scheduler/binding.dart:863:7)
[...]

PR #524, opened 2021-06-16
Disable desktop scrollbars

Piinks:
See https://github.com/flutter/gallery/issues/523

On desktop, scrollbars are applied automatically by the scroll behavior. Currently,  most vertical scroll views in the gallery app are all using the same primary scroll controller. This causes a crash when users try to interact with the scrollbars. This change will disable the automatic scrollbars to eliminate the crash while we audit and fix up all of the scrollables in the gallery that are using the same scroll controller.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -51,7 +51,8 @@ class GalleryApp extends StatelessWidget {
             // ScrollBehavior. This overrides that. All vertical scrollables in
             // the gallery need to be audited before enabling this feature,
             // see https://github.com/flutter/gallery/issues/523
-            scrollBehavior: const MaterialScrollBehavior().copyWith(scrollbars: false),
+            scrollBehavior:
+                const MaterialScrollBehavior().copyWith(scrollbars: false),
             restorationScopeId: 'rootGallery',
             title: 'Flutter Gallery',
             debugShowCheckedModeBanner: false,
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -47,6 +47,11 @@ class GalleryApp extends StatelessWidget {
       child: Builder(
         builder: (context) {
           return MaterialApp(
+            // By default on desktop, scrollbars are applied by the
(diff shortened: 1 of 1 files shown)


lib/main.dart
```diff
@@ -47,6 +47,11 @@ class GalleryApp extends StatelessWidget {
       child: Builder(
         builder: (context) {
           return MaterialApp(
+            // By default on desktop, scrollbars are applied by the
+            // ScrollBehavior. This overrides that. All vertical scrollables in
+            // the gallery need to be audited before enabling this feature,
+            // see https://github.com/flutter/gallery/issues/523
+            scrollBehavior: const MaterialScrollBehavior().copyWith(scrollbars: false),
```
dnfield, line 54: FWIW, this problem also exists in the legacy Flutter gallery.

I'm missing some context here to really understand why this is the right fix. In particular, why do certain scrollables need to set primary or not, when it used to "just work" that they didn't.

Piinks: > FWIW, this problem also exists in the legacy Flutter gallery.

Yeah, I need to track that down too.

> why do certain scrollables need to set primary or not, when it used to "just work" that they didn't.

Vertical scrollables automatically attach to the primary scroll controller if they aren't given a scroll controller. The auto-scrollbar stuff is new, and the gallery will need to be updated for it.  It 'just worked' before because there wasn't a widget in the tree enforcing that a certain scroll controller only has one position attached. Since the ScrollBehavior on desktop is adding scrollbars now, the gallery crashes because the when the user tries to interact with the scrollbar, the scrollbar tries to use the scroll controller to change the position. Since more than one position is attached, it breaks.

There is a migration guide here: https://flutter.dev/docs/release/breaking-changes/default-desktop-scrollbars

I am also updating the assertion in the framework because I do think it is a little too aggressive right now.

rami-a (approved):
LGTM

PR #528, opened 2021-06-28
Removed ButtonBar dependencies from shrine/login and the cards demo

HansMuller:
Replaced the two uses of ButtonBar, in shrine/login and the cards demo, with OverflowBar. The semantics and implementation of ButtonBar are problematic (for example see #53378) and the widget is no longer necessary.  

Similar changes were made to the original version of the gallery, see https://github.com/flutter/flutter/pull/85351.


rami-a (approved):
LGTM!

Issue #541, opened 2021-09-08
Shrine crashes when scrolling on desktop 


Piinks:
Shrine crashes when you scroll on desktop. This is caused by the scrollbars that are automatically applied on desktop platforms. There are scroll views in Shrine that are all using the same scroll controller, so when it tries to draw one scrollbar for all of those scroll views it crashes.

This is most likely due to the PrimaryScrollController. All **_vertical_** scroll views that have not been provided a scroll controller automatically attach to the PrimaryScrollController. If each scroll view is provided a unique scroll controller, the desktop scrollbars won't have an issue. I am opening a PR now to disable them and fix the crash until that work can be done.

Related: https://github.com/flutter/gallery/issues/523

PR #542, opened 2021-09-08
Remove desktop scrollbars from shrine

Piinks:
This prevents the Shrine app from crashing on desktop due to scrollbars and too many scroll views connected to the same scroll controller.
Fixes https://github.com/flutter/gallery/issues/541


dnfield (approved):
RSLGTM

PR #544, opened 2021-09-09
Make index.html compatible with older Safari

yjbanov:
The `??` syntax is not supported by Safari older than 13.


LongCatIsLooong (approved):
Thank you!

/cc @guidezpl gallery.flutter.dev isn't loading on older versions of safari and it does not seem to recognize "??".

guidezpl (approved):

guidezpl:
Thanks!

PR #556, opened 2021-11-04
Only use supported locales.

Hixie:
The previous implementation would sometimes select locales that were not actually supported. For example, on my desktop machine it selected "C_UTF-8".


rami-a (approved):
LGTM

Hixie:
thanks!

PR #564, opened 2021-12-09
Update feedback link

### LINKED ISSUE #536: Change your 'send feedback' link to stop pointing to Flutter
Your send feedback link on https://gallery.flutter.dev/#/ points to the Flutter github issues, incorrectly. I logged an issue here and was told it was the wrong forum. Change the link please

guidezpl:
Closes #536


rami-a (commented):
So we originally were thinking that users would find issues with Flutter widgets and use this to file them, but do Gallery issues make more sense instead?

guidezpl:
Yeah, it's actually been more the other way around. If issues are filed that are framework related we can transfer than to flutter/flutter

rami-a (approved):

PR #598, opened 2022-02-07
Remove unnecessary code from settings.gradle

blasten:
The content in settings.gradle is breaking the Flutter build.


HansMuller (approved):
RSLGTM

PR #596, opened 2022-02-05
Fix two null safety analysis issues and update goldens

devoncarew:
Fix two null safety analysis issues.

*List which issues are fixed by this PR. You must list at least one issue.*

- these analysis issues are currently breaking the Dart SDK's `flutter-analyze` CI bot (https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/flutter-analyze/15160/overview)

```
warning • The receiver can't be null, so the null-aware operator '?.' is unnecessary •
       lib/feature_discovery/feature_discovery.dart:224:34 • invalid_null_aware_operator
warning • The receiver can't be null, so the null-aware operator '?.' is unnecessary •
       lib/feature_discovery/feature_discovery.dart:240:36 • invalid_null_aware_operator
```

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #601, opened 2022-02-10
Fix dart null check

TestSmirk:
fix Error: Method 'addPostFrameCallback' cannot be called on 'SchedulerBinding?' because it is potentially null.


guidezpl:
This is expected for a little while. See https://github.com/flutter/gallery/pull/596

PR #659, opened 2022-04-25
Update firebase configuration to single project, multi-host

guidezpl:
The gallery was previously configured with two projects, one for prod, one for staging. Firebase now supports multiple hosts for a single project. Staging is now at https://gallery-flutter-staging.web.app/


clocksmith (approved):
LGTM

PR #691, opened 2022-05-18
allow picking a renderer using a query parameter

yjbanov:
Adding `?renderer=html` or `?renderer=canvaskit` in the URL will select the respective renderer instead of picking one automatically.


web/index.html
```diff
@@ -34,6 +34,17 @@
   </style>
   <link rel="manifest" href="manifest.json">
   <script>
+    let query = window.location.search;
+    if (query && query.indexOf('renderer=') !== -1) {
+      if (query.indexOf('renderer=html') !== -1) {
+        console.log('HTML renderer requested in the URL');
+        window.flutterWebRenderer = 'html';
+      } else {
```
ditman, line 42: I think this `else` should instead be an explicit check for `renderer=canvaskit` here (similar to the `html` above), otherwise forcing "canvaskit" is going to break "auto" mode in mobile iOS (when hitting gallery.flutter.dev)

hterkelsen (approved):
LGTM

web/index.html
ditman: Ahhh no, this is checking for renderer=anything in 38, nvm!

ditman (approved):

hterkelsen (approved):
LGTM

web/index.html
```diff
@@ -34,6 +34,15 @@
   </style>
   <link rel="manifest" href="manifest.json">
   <script>
+    let query = window.location.search;
+    if (query) {
+      let searchParams = new URLSearchParams(query);
```
ditman, line 39: URLSearchParams seems to support a nullable string passed in, so you could potentially just do:

```js
let searchParams = new URLSearchParams(window.location.search);
if (searchParams.has('renderer') {
  ...
}
```

yjbanov: Done.

guidezpl:
Staged, https://gallery-flutter-staging.web.app/?renderer=html and https://gallery-flutter-staging.web.app/?renderer=canvaskit work.

Issue #698, opened 2022-05-25
Replace `SelectableText` with a small number of `SelectableArea`s


guidezpl:
Need to check that selection works as expected with all nested apps (e.g. studies, as well as the settings screen.

See flutter/flutter#104547

AdiAr11:
could you please explain what needs to be done here?

olegzaim:
I have already started to work on that issue.

AdiAr11:
okay, no problem !

Roshani9731:
can I try this issue??

guidezpl:
@olegzaim has a PR out that fixes this issue, however 2 driver tests are failing. I'd welcome help in debugging those

guidezpl:
Now blocked on #104547

guidezpl:
Now that https://github.com/flutter/flutter/issues/104595 is closed, this issue becomes valid again in the short-/medium- term.  https://github.com/flutter/flutter/issues/104547 is still a long-term proposal.

PR #699, opened 2022-05-26
Migrate all SelectableText to SelectionArea.

olegzaim:
Replaced all SelectableText to SelectionArea.
Fixes https://github.com/flutter/gallery/issues/698


guidezpl (changes requested):
This is not quite what is required. Only ~5 `SelectionArea`s should be necessary, one per `MaterialApp`, and one for the setting page, each placed near the top of the widget tree.

Each `SelectableText` should be replaced by `Text`.

lib/studies/shrine/app.dart
```diff
@@ -164,7 +164,7 @@ class _ShrineAppState extends State<ShrineApp>
           initialRoute: ShrineApp.loginRoute,
           routes: {
             ShrineApp.loginRoute: (context) => const LoginPage(),
```
guidezpl, line 166: should work on the login page as well

lib/studies/shrine/app.dart
```diff
@@ -164,7 +164,7 @@ class _ShrineAppState extends State<ShrineApp>
...
-            ShrineApp.homeRoute: (context) => home,
+            ShrineApp.homeRoute: (context) => SelectionArea(child: home),
```
guidezpl, line 167: is there an `ApplyTextOptions` missing here? if not, can it be moved here for consistency?

olegzaim:
> * Rally text selection is invisible.

Hello, there is something wrong with the cursor for selection area because when I select all area it isn't invisible, but when I just tap and press it is stay invisible: 
- Single select with long press
long press for select
- Select all
select all

> * Demos don't have text selection

Is it enough or it's need to be selectable also inside demos?
home and demos select

guidezpl:
I made some small fixes for consistency. The only thing missing I think is adding selection for the demos (material, cupertino, etc.)

olegzaim:
> I made some small fixes for consistency. The only thing missing I think is adding selection for the demos (material, cupertino, etc.)

Okay, great thank you :)

guidezpl (approved):
Thanks for contributing!

guidezpl:
@chunhtai This change triggers errors that cause  benchmark and golden tests to fail, would appreciate you taking a look if possible.

guidezpl:
@olegzaim Could you take a look at the merge conflicts so that we can land this? Thanks!

guidezpl:
Realized you might only have access to @OlegZaimNLT, if so, please create a new PR cloning this branch

OlegZaimNLT:
Hey @guidezpl I can't merge because it says that I don't have permissions
Screenshot 2022-06-30 at 11 59 16
.

PR #719, opened 2022-06-24
Update deprecated parameters for styleFrom button methods

guidezpl:
Following https://github.com/flutter/flutter/pull/105291

Also add updated goldens to fix almost imperceptible rendering difference caused by upstream changes.


bwilkerson (commented):
The code changes lgtm, but I'm not familiar enough with Flutter to know whether the changes to the lock file and the golden files are reasonable.

darrenaustin (approved):
The code changes LGTM. I don't see the differences in the golden images so I don't think there is an issue with button colors changing here. Is there something else tripping the golden differences?

guidezpl:
Thanks! Golden changes are minute rendering differences due to upstream changes.

PR #722, opened 2022-06-27
Add disabled widgets to `material` and `cupertino` demos

guidezpl:
Improve the demos by showcasing disabled widgets.

This also fixes an issue with the toggleable widgets' states, and the header which appears when a data table item is selected.


lib/demos/cupertino/cupertino_slider_demo.dart
```diff
@@ -68,22 +70,29 @@ class _CupertinoSliderDemoState extends State<CupertinoSliderDemo>
                   ),
                 ],
               ),
+              // Disabled sliders
+              // TODO(x): See https://github.com/flutter/flutter/issues/106691
```
goderbauer, line 74: use a real name in the todo?

goderbauer (approved):
LGTM

Do we need tests for this change?

guidezpl:
Will just update existing tests.

PR #735, opened 2022-07-11
Enable scrollbars

### LINKED ISSUE #523: Auto-Scrollbars on Desktop are disabled

guidezpl:
Fixes #523

I can't repro the original issue, but out of caution, I've set `primary: false` on the secondary scrollable on the home screen.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -48,12 +48,6 @@ class GalleryApp extends StatelessWidget {
         builder: (context) {
           final options = GalleryOptions.of(context);
           return MaterialApp(
-            // By default on desktop, scrollbars are applied by the
-            // ScrollBehavior. This overrides that. All vertical scrollables in
-            // the gallery need to be audited before enabling this feature,
-            // see https://github.com/flutter/gallery/issues/523
-            scrollBehavior:
-                const MaterialScrollBehavior().copyWith(scrollbars: false),
             restorationScopeId: 'rootGallery',
             title: 'Flutter Gallery',
             debugShowCheckedModeBanner: false,
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -1431,7 +1431,7 @@
   "@demoCupertinoScrollbarTitle": {
     "description": "Title for the cupertino scrollbar demo."
   },
-  "demoCupertinoScrollbarSubtitle": "iOS style scrollbar",
+  "demoCupertinoScrollbarSubtitle": "iOS-style scrollbar",
   "@demoCupertinoScrollbarSubtitle": {
     "description": "Subtitle for the cupertino scrollbar demo."
   },
--- a/lib/l10n/intl_en_US.xml
+++ b/lib/l10n/intl_en_US.xml
@@ -1324,7 +1324,7 @@
   <string
     name="demoCupertinoScrollbarSubtitle"
     description="Subtitle for the cupertino scrollbar demo."
-    >iOS style scrollbar</string>
+    >iOS-style scrollbar</string>
   <string
     name="demoCupertinoScrollbarDescription"
     description="Description for the cupertino scrollbar demo."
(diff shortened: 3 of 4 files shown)


lib/main.dart
```diff
@@ -48,12 +48,6 @@ class GalleryApp extends StatelessWidget {
         builder: (context) {
           final options = GalleryOptions.of(context);
           return MaterialApp(
-            // By default on desktop, scrollbars are applied by the
-            // ScrollBehavior. This overrides that. All vertical scrollables in
-            // the gallery need to be audited before enabling this feature,
-            // see https://github.com/flutter/gallery/issues/523
-            scrollBehavior:
-                const MaterialScrollBehavior().copyWith(scrollbars: false),
```
Piinks, line 56: I think the error presented itself in other demo apps in the gallery as well. We recently changed the default PrimaryScrollController behavior (https://github.com/flutter/flutter/pull/102099) to better avoid that error in general, so it may be ok.

I think the Shrine app was one that also had this issue.

Piinks (commented):
Does the gallery write tests?

lib/main.dart
guidezpl: Ah I see. Can't repro on Shrine anymore.

Piinks (approved):
LGTM!

PR #737, opened 2022-07-13
Add cookie banner to web

### LINKED ISSUE #718: Add cookie banner
```
This site uses cookies from Google to deliver its services and to analyze traffic.

  [Learn more.](https://policies.google.com/technologies/cookies)    Ok, Got it.
```

Prereq for landing https://github.com/flutter/gallery/pull/676

guidezpl:
This PR adds a mechanism for acknowledging the use of cookies through a banner. The same styling as on `flutter.dev` is used, and it displays on all navigateable pages. The banner appears if the cookie isn't set, and the cookie can be set by pressing "Okay".

Screenshot 2022-08-25 at 10 22 51

Fixes #718 

Took the opportunity to reorganize the `index.html` file for clarity.


guidezpl:
Now that I think about it, it might be simpler to avoid using `ScaffoldMessenger` altogether and just have it display atop every page from `routes.dart`?

Piinks (commented):
> WIP because the banner currently appears on all screens except demos. DemoPage already uses scaffolds, but an error is raised and the banner doesn't display

By default with nested Scaffolds, only the topmost scaffold will present.

I am not very familiar with the Gallery app, but I would check the context from where you are calling to display the banner. The context of the call needs to be below the ScaffoldMessenger, there must also be a Scaffold below it.
I would check and see in the case of StudyWrapper, where is the ScaffoldMessenger you are expecting to handle this? There are some manually placed ScaffoldMessengers in the Gallery as you found in https://github.com/flutter/gallery/pull/377, this is because of the Gallery's Flutter app inside of a Flutter app demo nature.

The other thought I have is that since you are calling to show the banner in a postframe callback, the Scaffold may not be there anymore because the page changed and it unregistered from the ScaffoldMessenger.

guidezpl:
I went with using HTML + javascript simply instead, which is a bit more suitable in this case.

esouthren (approved):
LGTM!

PR #747, opened 2022-08-05
Enable spacebar to scroll on the homepage

guidezpl:
https://github.com/flutter/flutter/issues/108735


Piinks (commented):
I am not familiar with this app, why are there 2 `primary`? Are both present on the home screen at the same time? If so, this will probably cause a crash.
Is there test coverage for this change?

guidezpl:
No, it's two different layouts. Manually tested, I don't know how to test keyboard entry

Piinks (commented):
If it is untested, what prevents this from regressing in the future?

guidezpl:
Praying? 😄

I tried adding a test but can't get it to pass: https://github.com/flutter/gallery/blob/78900fd40ce12e7bebeed888784fdf8417798a37/test_integration/interactions_test.dart. I tried key up + key down, with/without `PhysicalKey`, etc.

Piinks (commented):
Hmm, without testing, the Gallery is going to just continue to experience code rot like this. It really should be tested.
Maybe instead of an integration test, you can just use a widget test? Here is a very similar test from the framework: https://github.com/flutter/flutter/blob/ba5181b7885340bacd370eac779c319dfd82670e/packages/flutter/test/widgets/shortcuts_test.dart#L984

guidezpl:
I spent a good amount of time trying to get the test to work, and even checking the multiple `ScrollController`s found, the the pixel value is always 0. I give up 🤷‍♂️

PR #754, opened 2022-08-23
Regular maintenance

guidezpl:
- update mac & iOS runners
- update packages, notably `google_fonts` to 3.0.1
- replace deprecated M2 `TextStyle` names
- regen some Android build files
  - fix build issues
  - bump Kotlin and Gradle versions
- update GitHub workflow actions to use latest runners, and setup Java 11 which is now required by Gradle for building Android
- update goldens due to imperceptible rendering changes
- update translations (8 more strings coming in ~2 weeks)


guidezpl:
@godofredoc I've updated our workflows to always use the latest images, so could you update the required statuses accordingly?

godofredoc:
> @godofredoc I've updated our workflows to always use the latest images, so could you update the required statuses accordingly?

Is it possible to make the test names generic? as it is the tests will be broken every time the version changes behind the scenes

guidezpl:
Using `x-latest` makes them generic, so this should be the last time they require modifying.

guidezpl:
Whoops, forgot one, should be good now.

godofredoc:
> Whoops, forgot one, should be good now.

Updated!

jason-simmons (approved):

PR #780, opened 2022-09-13
Permission fix for scorecards 2.0.3

guidezpl:
Following https://github.com/flutter/flutter/pull/111441 by example.


godofredoc (approved):
Thanks for the fix!

PR #783, opened 2022-09-17
Fixing unchecked use of a nullable value

Mohamed-7018:
The method 'insert' can't be unconditionally invoked because the receiver can be 'null'.


Mohamed-7018:
I signed it!

PR #784, opened 2022-09-17
Fixing unchecked use of nullable check

Mohamed-7018:
"The method 'insert' can't be unconditionally invoked because the receiver can be 'null'


Mohamed-7018:
I signed in!

guidezpl:
Please see the README's troubleshooting section

PR #797, opened 2022-10-05
Fix new syntax for escaping with single quotes

thkim1011:
We're adding a new syntax to support escaping in ICU MessageFormat via single quotes. According to official documentation, the behavior is that any characters within single quotes will not be parsed and to use the single quote character in a string, you need to use a pair of consecutive single quotes. See [here](https://unicode-org.github.io/icu/userguide/format_parse/messages/#quotingescaping).

The only changes in this PR is
* Change all instances of `'` to `''`.
* Change `{per cent}` to `{percent}` in some English translation, since the former is not valid syntax.

I checked that `flutter gen-l10n` runs when using the branch in the PR below.

Related PR: https://github.com/flutter/flutter/pull/112390.


guidezpl (approved):
Golden test failures are unrelated

PR #802, opened 2022-10-13
Revert "Fix new syntax for escaping with single quotes"

thkim1011:
Reverts flutter/gallery#797

My bad, I should've closed this PR last week. I noticed that you can't just change the *.arb files since they're generated from TC. I'm looking into a different way to fix the issues.


guidezpl:
No worries, I also forgot about this.

guidezpl (approved):

guidezpl:
Consider looking at `third_party/dart/flutter_gallery/push_l10n/lib/src/push_l10n.dart`

PR #856, opened 2023-01-03
Replace deprecated ScrollPhysics.tolerance with ScrollPhysics.toleranceFor

goderbauer:
Deprecation clean-up for https://github.com/flutter/flutter/pull/117916.

To be submitted after that CL.

PR #857, opened 2023-01-09
Fixed crane build issue

Vishnuv2311:
## Fixed build issue

-  ~~Fixed tolerance function not found issue.~~
-  Fixed eat image not found issue.


guidezpl (approved):
That change is expected (https://github.com/flutter/gallery/pull/856) so I've reverted that file. But thanks for fixing the missing image

PR #864, opened 2023-01-13
Perform regular maintenance

guidezpl:
- Update dependencies, specifically loosen the constraint for `intl`
- Update iOS / macOS Podfiles
- Bump SDK lower bound

Enables https://github.com/flutter/flutter/issues/117163


esouthren (approved):
LGTM

PR #900, opened 2023-02-13
Suppress l10n warnings

### LINKED ISSUE #893: Address l10n errors
When running the gallery, a slew of these errors appear 
```
[intl_zu.arb:craneMinutes] ICU Syntax Warning: The plural part specified below is overridden by a later plural part.
  {minutes,plural, =1{1m}one{{minutes}m}other{{minutes}m}}`
```

They're not actionable but we should figure out what they are and how to resolve them.

cc @HansMuller who knows about this topic?

thkim1011:
## Description
This change suppresses l10n warnings. Currently the only warning implemented in the `gen-l10n` tool is to warn users if they have a repeated plural case, since this renders one of the plural case useless. However, it seems that Google distinguishes between the `=1` and `one` case, but these are interpreted to be the same thing in ICU's messageFormat syntax.

## Issues
Fixes #893.

### DIFF
--- a/l10n.yaml
+++ b/l10n.yaml
@@ -4,3 +4,4 @@ output-class: GalleryLocalizations
 preferred-supported-locales:
   - en
 use-deferred-loading: true
+suppress-warnings: true


guidezpl (approved):
Nice, thank you

PR #906, opened 2023-02-20
const was wrongly placed

oiiakm:
- const was wrongly placed

- const was missing


guidezpl:
The gallery targets the flutter's main channel, so this is expected.

Issue #904, opened 2023-02-17
Many routes are unconditionally using the foldable navigation strategy


jonahwilliams:
### Existing issue?

- [X] I checked the [existing issues](https://github.com/flutter/gallery/issues)

### What happened?

See also: https://github.com/flutter/flutter/issues/120937

If openInSecondScreen is true, we unconditionally use a TwoPaneScreen. This is implemented as a OverlayRoute, which is typically used for popups and modal menus, so it does not hide the contents of the previous route. This leads to dramatically reduced rendering performance and no page transition, even if on a non-foldable device. 

Instead the hinge check should be hoisted into route selection so that we always use a MaterialPageRoute on non-foldable devices

https://github.com/flutter/gallery/blob/main/lib/routes.dart#L133

PR #909, opened 2023-02-28
dont use TwoPanePageRoute if there is no hinge

jonahwilliams:
Fixes https://github.com/flutter/gallery/issues/904

See also: https://github.com/flutter/flutter/issues/120937

The TwoPanePageRoute doesn't use a ModalRoute, leaving the previous route fully active. Additionally, there is no page transition on these route transitions.


lib/routes.dart
```diff
@@ -118,7 +118,8 @@ class RouteConfiguration {
   /// route. Set it on the [MaterialApp.onGenerateRoute] or
   /// [WidgetsApp.onGenerateRoute] to make use of the [paths] for route
   /// matching.
-  static Route<dynamic>? onGenerateRoute(RouteSettings settings) {
+  static Route<dynamic>? onGenerateRoute(
+      RouteSettings settings, bool hasHinge) {
```
guidezpl, line 122: nit: trailing comma

PR #935, opened 2023-04-26
Fixed small typos on windows deployment yaml.

azchohfi:
## Description

Very small typo fix. Secret name was mistyped, and the dot (".") is not required, as the Store CLI uses the current directory by default.


azchohfi:
Just noticed that this build is also using dart pub get, which is wrong. I'll update that too.

yaakovschectman (approved):
LGTM

PR #936, opened 2023-04-26
Fix crash, goldens, demo status bar icon color

guidezpl:
## Description

For 2.10.1 release:

1. Fix a common crash when opening a demo for the first time, caused by feature discovery. `shared_preferences` had undesireable asynchronous behavior, so I replaced it with `get_storage`.

2. I also wanted to fix the goldens test, but found the `Positioned` widget around the settings icon was unnecessarily adding  padding, so removed it.

image

3. Fix the demo status bar icon color in light mode to use the correct `SystemUiOverlayStyle`.

## Tests

Tested manually

## Issues
Fixes Firebase-detected crash


lib/feature_discovery/overlay.dart
```diff
@@ -323,8 +322,10 @@ class TapTarget extends StatelessWidget {
             child: Container(
               height: radius * 2,
               width: radius * 2,
-              decoration: const BoxDecoration(
-                color: Colors.white,
+              decoration: BoxDecoration(
+                color: Theme.of(context).brightness == Brightness.dark
```
esouthren, line 326: Consider extracting `theme` to a variable here

lib/pages/demo.dart
```diff
@@ -247,6 +220,8 @@ class _GalleryDemoPageState extends State<GalleryDemoPage>
     final localizations = GalleryLocalizations.of(context)!;
 
     final appBar = AppBar(
+      systemOverlayStyle:
+          GalleryOptions.of(context).resolvedSystemUiOverlayStyle(),
```
esouthren, line 224: Same here for `galleryOptions` (extraction)

esouthren (changes requested):
LGTM with some minor comments

esouthren (approved):

PR #949, opened 2023-05-08
Regenerate localizations

thkim1011:
## Description

Updating localizations with updated `push_l10n.dart`. A side effect of parsing through the translations and removing redundant plural parts is that all plural messages are recreated without any whitespace, causing a larger diff for this PR than it actually should be.

## Tests
Checked that `flutter gen-l10n` runs without errors or warnings.
## Issues
Fixes https://github.com/flutter/flutter/issues/108918 since we no longer need to suppress warnings.

---

>**Note**: Please find the development and releasing instructions at https://github.com/flutter/gallery#development


guidezpl (approved):
nice

PR #952, opened 2023-05-27
Reply: Remove unnecessary inboxKey from AdaptiveNav

sjoonb:
## Description

If the intent is to change the state of the MailboxBody widget based on a change in the selected destination, I believe that the Provider package already handles this efficiently.


guidezpl (approved):
Thanks!

Issue #967, opened 2023-06-09
Rotation doesn't work in 2D transformations demo


rorystephenson:
### Existing issue?

- [X] I checked the [existing issues](https://github.com/flutter/gallery/issues)

## Expected vs actual result:

Two finger rotate should rotate the the scene (as described when tapping the info button), it does not.

Tested on:
1. Latest master (d77920b4ced4a105ad35659fbe3958800d418fb9) with flutter master (flutter doctor output below).
2. [Flutter Gallery Android app](https://play.google.com/store/apps/details?id=io.flutter.demo.gallery) using real Android device (according to the 'About Flutter Gallery' dialog it is version 2.9.2).

## Steps to reproduce:
1. Open 2D transformations demo (Styles & Transformations > 2D Transformations).
1. Try to two-finger rotate.

### Relevant log output

```shell
[✓] Flutter (Channel master, 3.12.0-3.0.pre.31, on macOS 12.5.1 21G83 darwin-arm64, locale en-NZ)
    • Flutter version 3.12.0-3.0.pre.31 on channel master at /Users/rorystephenson/sdk/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 192b06405b (49 minutes ago), 2023-06-09 11:04:07 +0300
    • Engine revision 8f9e608d39
    • Dart version 3.1.0 (build 3.1.0-180.0.dev)
    • DevTools version 2.24.0

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    • Android SDK at /Users/rorystephenson/sdk/android
    • Platform android-33, build-tools 33.0.1
    • ANDROID_HOME = /Users/rorystephenson/sdk/android
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
[...]

PR #968, opened 2023-06-09
Remove rotation from Transformation demo description since it's not implemented

justinmc:
## Description

The transformation demo does not include rotation, but the description and subtitle says it does.  This PR removes mention of rotation.

Rotation is not currently supported by InteractiveViewer under the hood, though someday I'd like it to!

@guidezpl What's the process for updating an existing string like this?  I couldn't figure it out by reading the [l10n readme](https://github.com/flutter/gallery/tree/main/lib/l10n#readme).

## Tests

None.

## Issues
Fixes https://github.com/flutter/gallery/issues/967


guidezpl (approved):
Did you see https://github.com/flutter/gallery/tree/main/lib/l10n#obtaining-translations-for-flutter-org-members and `third_party/dart/flutter_gallery/README.md`?

justinmc:
Yes, but it didn't seem to work for me.  Sorry, I'm probably just inexperienced with l10n stuff.

Running `flutter pub run grinder l10n` overwrites intl_en_US.xml with the old strings.

Running `flutter gen-l10n` gives:

```
Because l10n.yaml exists, the options defined there will be used instead.
To use the command line arguments, delete the l10n.yaml file in the Flutter project.
``

And doesn't seem to make any changes.

**Comment by guidezpl** [comment]
Me too, it's been a while. So it's not super clear (feel free to update the instructions), but the `arb` files serve as source, and in particular `intl_en.arb` serves as sources for the singular `intl_en.US.xml` file. So once you add a string to the former, `flutter pub run grinder l10n` updates the latter, which can then be imported into Google, translated, etc.

**Comment by justinmc** [comment]
@guidezpl Let me know if the PR looks good now to the best of your knowledge.  I update the `arb` file as well, and I added a section in the README about udpating messages as far as I understand it.

**Review by guidezpl — APPROVED** [review]
Yep, 
```
  _    ___ _____ __  __ 
 | |  / __|_   _|  \/  |
 | |_| (_ | | | | |\/| |
 |____\___| |_| |_|  |_|
                        
```

justinmc:
@guidezpl It looks like goldens are failing.  I guess this text is included in some golden images?  Can you confirm that the changes are expected?  When I run the goldens locally with `flutter test test_goldens`, I get failures even on main.

If you confirm it's ok, it looks like I can update the goldens with `flutter test --update-goldens test_goldens`.

guidezpl:
The goldens at https://github.com/flutter/gallery/actions/runs/5246680988 look good to me. Typically this is because some change hit Flutter @ main, and we can just update the goldens from that `goldens.zip` file. It's easier to use that since it's was CI tests against

PR #996, opened 2023-08-22
Shrink image sizes of attachments in Reply app.

jonahwilliams:
During the investigation of https://github.com/flutter/flutter/issues/132690 , we discovered that the Reply image previews are the fully size 2000x2000 images. The decoding of these images is putting the device under heavy strain which is making it hard to diagnose other performance issues.


zanderso (approved):

Renzo-Olivares (approved):

jonahwilliams:
So how do I update these goldens? I don't see the download button in the Github UI like https://github.com/flutter/gallery/blob/main/test_goldens/README.md shows

zanderso:
I see links under the step "Upload goldens if tests fail", but I get auth errors(?) when I click on them. Not sure who is on the right ACLs for this repo. Maybe @HansMuller or @goderbauer know?

goderbauer:
I also get auth errors. @guidezpl would probably know how to do this.

PR #1023, opened 2023-10-25
Delete firebase dependencies to simplify framework roll.

jonahwilliams:
We're having trouble updating the flutter gallery to head for flutter integration testing. Since we're not updating the gallery any further, I believe we should simplify the problem by removing the firebase depenencies, which are complicating the native build.

This would ultimately be required to merge this gallery into flutter/flutter anyway.


goderbauer:
This sounds reasonable to me. @guidezpl or @HansMuller can you think of any reasons to keep these dependencies around now that Gallery is no longer an "active" showcase?

guidezpl:
The purpose of this was two-fold: to get information about crashes and usage analytics: 340k web hits in the last year and 26k installs on Android. Part of the reason is that it's still featured prominently on the Flutter docs website. Removing Firebase would mean losing this information.

Alternatively, would widening the firebase constraints help rolling it in to flutter/flutter?

jonahwilliams:
Are we still doing deploys of this application then? It was my understanding there would be no further updates

jonahwilliams:
Friendly ping @guidezpl

jonahwilliams:
(IF there are going to be further updates then I'm fine leaving this as is, we can try loosening the constraints and removing the lock files)

domesticmouse (approved):

PR #1046, opened 2023-11-07
Update intl_et.arb

villu164:
**näide.com** is not the same as **example.com**

Read more - https://secret.näide.com/

## Summary
Localization was fixed, to use example.com domain not a private domain

## Motivation
Security related. Not an urgent(probably seen as a no-issue by many) security issue, but could/should trigger the thought process of the same "example.com" email in other languages

## Testing
Compare the output of
[https://example.com/](https://youtu.be/dQw4w9WgXcQ?Seriously+Watch+Where+You+Are+Clicking)
https://näide.com/
https://secret.example.com/
https://secret.näide.com/
To verify that they are not the same


villu164:
Any updates?

guidezpl:
Hi, we are turning down the Gallery, and I therefore must close this PR.

